Issue #6, opened 2017-04-20
Coding Style Guide


andygrunwald:
Working in a team on the same codebase might be tricky.
One step to be able to read and understand code of others quite fast is to follow the same coding style guide.
With this you don't have to spent time thinking about the code structure, but can focus on the business logic.

## Suggestion

My proposal is to follow a common standard in the php world and use

* [PSR-1: Basic Coding Standard](http://www.php-fig.org/psr/psr-1/)
* [PSR-2: Coding Style Guide](http://www.php-fig.org/psr/psr-2/)

for your coding style guide. Bigger projects follow those rules as well.

### Do we need to check those style guides manually?

No we don't. Because it is a common and well accepted standard, there are tools out there to support this. One example is the [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer). The CodeSniffer has predefined rule sets for PSR-1 and PSR-2.
So we are ready to use it directly.

### But who formats the complete codebase according the styles? It is a lot of work

Thanks to computers and standards, we don't have to do it.
With the tool [PHP Coding Standards Fixer](http://cs.sensiolabs.org/) we are able to reformat the code according to the guidelines.

## How do we come to a conclusion to solve this ticket?

In my opinion this has to be a decision of the main contributors. I would encourage them to post their **+1 for those cgls** or a **-1 for stay with the current state** here.

jannaahs:
I had some trouble reading the code at the current state.
In order to become a main contributor I voted with a yes for the Coding Style Guide.

andygrunwald:
In #10 i applied the first PR for module "about".
If @M4LuZ will agree here as well, i will take care about all modules and the rest of the code base in single commits.

But be aware: If you have uncomitted / unmerged changes to the existing codebase / modules, you might get conflicts (because of reformatting).

JochenJung:
I thought the lansuite developer documentation already had the coding style guidelines listed, but the only thing I found was a naming guideline for database tables and fields:

http://lansuite.orgapage.de/index.php?mod=wiki&action=show&name=Database%20Design%20Guideline

andygrunwald:
3 people voted for it (@jannaahs, @skoths and me). @JochenJung said nothing against it. So i would say we try this.

I created all PRs for all modules.
If those are merged, i would continue with other parts of the application.

@jannaahs Feel free to review them and merge them.

jannaahs:
The new pull request only formatted the code.
[...]

PR #10, opened 2017-04-20
Apply coding style guide PSR 1 and PSR 2 to ./modules/about

andygrunwald:
This is the first try for #6 

I applied [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) and the [phpcbf](https://github.com/squizlabs/PHP_CodeSniffer) script for PSR 1 and PSR 2 to only the about module.

Be aware: Those automated tools don't fix everything, but ~90 or 95% percent. I think this is a good basement to continue.


jannaahs:
Only code formatting. It seems to be all right. Can I merge the pull request?

andygrunwald:
@jannaahs Go for it :)

jannaahs:
@andygrunwald I did it! Wuhu!

PR #62, opened 2017-04-23
#6: Apply coding style guide PSR 1 and PSR 2 to design/ folder (with phpcbf)

andygrunwald:
Ticket: #6 

I applied [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) and the [phpcbf](https://github.com/squizlabs/PHP_CodeSniffer) script for PSR 1 and PSR 2 to only the about module.

Be aware: Those automated tools don't fix everything, but ~90 or 95% percent. I think this is a good basement to continue.

### DIFF
--- a/design/beamer/style.css
+++ b/design/beamer/style.css
@@ -16,333 +16,441 @@ name {
 
 /* Main-Template */
 body {
-  margin: 0px 0px 0px 0px;    padding: 0px 0px 0px 0px;
-  text-align: left;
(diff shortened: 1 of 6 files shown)


andygrunwald:
I merge this pull request here now, because it is only CGL adjustments. No functional changes.
This was kind of approved in #6.

If something breaks, we can revert the commit later on again.

Issue #68, opened 2017-04-26
Documentation of classes and interfaces


andygrunwald:
@byte666 wrote in Ticket #6:

> What about documenation of the classes? I think thats also a important part of styling guides. In the past I have added many doxygen/phpdocumentor compartible dokumentation. Encapsulate the stuff in classes and documentation was (and is still) a base part to get a stable Lansuite. One of my last actions was to create more classes with clear interfaces and expand the pluginsystem. And there you need a good interface description.

I created a new ticket to discuss this (suggested by @M4LuZ in #6).

My 2 cents on this topic:
I agree with @byte666 here, but i think you are mixing several topics together:

### Class / Interface / Application-Architecture

You mentioned that during your last actions you created more classes with a clear interface. That is really great and necessary. Especially when we want to reuse the code. And good interface design is really difficult. Mostly those kind of action is bundled with a usage of a certain Design Pattern.
So: Thanks for the effort! I hope you are able to contribute more of those to lansuite :)
Would love to see this

### Documentation of Classes / Interfaces

Of course, one part of a style guide or one part of software development at a certain level is the documentation of a class / interface including their single methods and properties.
This is necessary and should be done.
The most common way in PHP is due to the usage of doc blocks like

```php
<?php

/**
 * Interface Foo is doing X Y and responsible for Z
 */
interface Foo {

    /**
     * My function is calculating the length of $bar and addint $baz on top.
     * Why? Because someone said so.
     * 
     * @param string $bar
     * @param int $baz
     * @return string
     */
    public function myFunction($bar, $baz);
}
```

And with the usage of annotations like `@param`, `@return`, etc.
Nearly every IDE or standard tool is able to understand those.

### Documentation of the architecture

To document the bigger architecture and the correct usage and combination of multiple interfaces and classes is a different part of the documentation. But important as well.
But i would prefer to put this into a more "prosa" text instead of doc blocks.

This kind of documentation is mostly part of a design document or part of a guide like "How to write a plugin".
One example could be the [Best Practice of the Bundle guide from Symfony](https://symfony.com/doc/current/bundles/best_practices.html).

### Format of the documentation

I mentioned different types of documentation: Code / interface documentation and architecture / design documentation.

#### Code / interface documentation

This part should be part of the code and next to the class as doc blocks. If someone changes the class, the docs are changed immediately. Even a proper IDE marks this via a red underline.
[...]

PR #69, opened 2017-04-26
Issue #68: Drop pre generated docs/ folder, because we are able to generate those for every release

andygrunwald:
See #68 for detail reasons.


M4LuZ:
Wouldn't it make sense to have  at least a note somewhere how to re-create the docs ready first?
Or did I just miss that?

andygrunwald:
@M4LuZ Right. This make sense :)

I added a note to the README on how to generate the HTML Version of the API.
PDF and CHM is not possible anymore, because my example is using the [phpDocumentor](https://www.phpdoc.org/) in Version 2.

Does this fit for you? If yes, feel free to merge it!

andygrunwald:
Any update here @M4LuZ ?

@jannaahs What is your opinion?

jannaahs:
I am fine with the note in the README.

M4LuZ:
Hi,

fine by me.
Regarding PDF and CHM:
Former would be good, I don't think we need the later.
But I'm quite sure that phpDocumentor is able to export to PDF.
Even if not, I think the syntax is also parseable by Doxygen, so that shouldn't be an issue either way

PR #70, opened 2017-04-26
Issue #3: Add PHP 7 as a requirement into README

andygrunwald:
See #3


M4LuZ:
So, if I see this correctly:
We haven't completely determinded, that we are compatible with PHP 7 yet, 
but say it is a requirement already.
This should be the other way around, shouldn't it? ;)

andygrunwald:
In my opinion this is a kind of chicken egg problem:

1. We want to be compatible with PHP7 and drop support for PHP <7. If we start now using PHP7 features and the README mentioned not the requirements for PHP7, people might using it and claim "Why are you not running on PHP 5.6"?
2. We said now we ned PHP7 as a requirement and are not 100% running on PHP7.

So i see this as a "in development phase" thing. We say we need PHP7 already and develop those things in the next time.
Does this make sense @M4LuZ ?

PR #77, opened 2017-05-19
Removed mysql_ functionality in favor of mysqli

andygrunwald:
mysql is deprecated since some time.
mysqli or PDO is the successor our of it.

In this PR we remove the mysql_* functionality at all.
Further more we remove a function form db class called `field_exist` that is a) not existant in mysqli and b) not in use in lansuite at all.


inc/classes/class_db_mysql.php
```diff
@@ -2,48 +2,30 @@
...
+    public $record = array();
     public $success = false;
     public $count_query = 0;
     public $errors = '';
     public $errorsFound = 0;
     public $connectfailure = 0;  //0= no error, 1=connection error, 2=database error
     public $QueryArgs = array();
 
-  // Construktor
-    public function __construct()
```
M4LuZ: Removing the constructor also removes the theck that the module for mysql(i) is loaded.
Not a huge issue by itself, it will just fail elsewhere and we need to make sure that the error is handled properly

andygrunwald: This is true. There are two options:
1. Introduce the constructor again and implement the HTML-Error via echo
2. Document the needed mysqli extension and let it fail with a Fatal error: undefined function

I would go for option 2. Because in the installation process has a check already in it if mysqli is supported. Further more for existing installations i would implement a small /health.php check or whatever that checks all requirements at once.

I would not suggest option 1, because if we want to make lansuite templatable / translatable, we need access to the template and translation system into the database class. This makes it hard to test.
An alternative would be throwing an exception in the constructor. And handle it everywhere.
But i would go for a global system check instead of firing this check at _every_ page call.
What do you prefer?

M4LuZ: As said I don't see a large problem with this. 
Only case we should avoid at all costs is that we leak the call parameters of mysqli_connect() in case initialisation fails. If that doesn't happen, then number 2 is perfectly fine.

M4LuZ: Gave it a try, no credentials are leaked if the connection attempt fails.
Error message thrown is:

`Das MySQL-PHP Modul ist nicht geladen. Bitte füge die mysql.so Erweiterung zur php.ini hinzu und restarte den Webserver neu. Lansuite wird abgebrochen
Fatal error: Uncaught Error: Call to undefined function mysql_connect() in C:\xampp\htdocs\lansuite\inc\classes\class_db_mysql.php:65 Stack trace: #0 C:\xampp\htdocs\lansuite\index.php(210): db->connect(1) #1 {main} thrown in C:\xampp\htdocs\lansuite\inc\classes\class_db_mysql.php on line 65`

PR #78, opened 2017-05-20
Remove the set of "error_reporting"

andygrunwald:
error_reporting should not be set in the app itself.
It should be set in the server configuration. The server / lansuite admin should evaluate what errors to show to the users and what not.

The issue here: If someone wants to get way more detailed error_reporting and is configured by the server, the app will overwrite it.

### DIFF
--- a/index.php
+++ b/index.php
@@ -1,17 +1,10 @@
 <?php
-### Set Error Reporting & INI-Settings
+// Set error_reporting properly
+// Right now we depend on it. This will change in future.
+error_reporting(E_ALL ^ E_NOTICE ^ E_DEPRECATED ^ E_STRICT);
 
-if (defined('E_STRICT')) {
-    error_reporting(E_ALL ^ E_NOTICE ^ E_DEPRECATED ^ E_STRICT);
-} // Will work for PHP >= 5.3
-elseif (defined('E_DEPRECATED')) {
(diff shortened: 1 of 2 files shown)


M4LuZ:
While I do agree that it is contra-productive to silently modify the error level, removing it also poses a problem:.
Lansuite expects a certain error level at the moment to function correctly.
Take that out and you add an undocumented implicit dependency on the php configuration.

Webhoster do not necessarily grant access to the php.ini used for the hosting instance, leaving you only with .htaccess or error_reporting() in the source code to set it.

And what text file you modify (php.ini or index.php) doesn't make that much of a difference, does it?

What we could do:
* Add a configuration option to change the error reporting in lansuite directly
* Add a check for the installer to check the running config error level against the expected value (but this does not cover existing installations)
* Move the error level setting to .htaccess (but that requires the webserver to allow modifications of php values e.g. Apache AllowOverride to be set correctly)

andygrunwald:
Good point @M4LuZ.
What i did right now:
- Removed error_reporting check from installation check
- Added the LanSuite default `E_ALL ^ E_NOTICE ^ E_DEPRECATED ^ E_STRICT` to index.php (so the installation will set it

I wouldn't go for the `.htaccess` solution, because a Nginx webserver don't respect such a `.htaccess` at all.

So i implemented the error_reporting back into the code and continue working on be able to remove these error_reporting at all :) Step by step :)

Please check the PR again. Thx

PR #79, opened 2017-05-20
Remove hardcoded set date_default_timezone_set to Europe/Berlin

andygrunwald:
The is a hardcoded `date_default_timezone_set('Europe/Berlin')` in the start application.
This makes it hard for other people outside of the "german timezone" to use this app like US, Portugal or other countries.

This setting should be setted by the env in the server (php.ini)


M4LuZ:
No issues here.
I recently had an install of 5.6 shipped without any timezone set per default, but we should be fine with PHP 7

PR #81, opened 2017-05-20
Introduced composer and load Smarty from it

andygrunwald:
[Composer](https://getcomposer.org/) is quite the standard solution to dependency management with PHP.
This PR introduces composer to LanSuite and load Smarty from package management instead of included itself.


andygrunwald:
@M4LuZ Any opinion / feedback here?

M4LuZ:
Hi Andy,

my opinion that this is fine, but I would suggest to merge it when we have the build structure ready so that we can provide a complete package for installation.

lxp:
I think our top priority should be to get rid of dependencies currently included in the LanSuite repository.
That's why, I would strongly opt for merging the pull request right now.
We can still make LanSuite a full composer package later. It might need a lot of refactoring and just blocks this pull request unnecessarily.

andygrunwald:
I 100% agree @lxp here.
@M4LuZ Are you fine with it?

M4LuZ:
@lxp: this wasn't about making it full-composer-ready, but to ensure that we are able to release full versions for end-users when we remove the smarty libraries.

I feel that we should do a release very soon as there are a lot of improvements between the last official LS release and now. 
But I just realized that this is not dependend on the HEAD code state, as we can prepare a release branch and pull only non function-breaking changes there.

Nevertheless it would be good to have a build script in Travis that allows us to provide a full package for any user. 
...given that we use Travis as CI-system. 
That discussion did not come up so far, but I would say that it is fine if it get the job done ;)

lxp:
@M4LuZ Ah, I didn't know that you are expecting a release soon, but it makes sense to show that the project is alive.

Sorry, that I just assumed Travis is okay for everyone. I know it from another open source project. Travis had it's problems there, especially because the project has a huge test framework that takes a long time to run. Otherwise, it worked as expected and is free for open source projects. We can still switch to another CI-system, if there are better options.

Maybe we should open a new issue for it, but what are you expecting the "full package" to contain? Isn't it just the full repository source?

M4LuZ:
@lxp I'm not really expecting one but rather proposing to make one happen ;)

As said I'm fine with Travis as it seems to get the job done and does provide quite a few benefits:
* Github integration out of the box
* At least some level of experience existing
* We don't need to worry about the hosting
So why argue if it seems to be the best choice? ;)

By full package I mean a ready-to-go installation source as previously provided.
Meaning the full source code with all dependencies, so that any user with just basic webhosting can get it up and running without resorting to a second machine with shell access to prepare the installation.

PR #84, opened 2017-07-05
Update ls_conf.php

cruelviper:
on:
index.php?mod=install&action=ls_conf
i get the error: 
Fehler: 
PHP Warning: file_exists(): open_basedir restriction in effect. File(design/style.css/design.xml) is not within the allowed path(s): (/var/www/vhosts/*/:/tmp/) in /var/www/vhosts/*/modules/install/ls_conf.php on line 72


andygrunwald:
Looks good! Thanks @cruelviper

PR #87, opened 2017-07-14
Fix syntax error in modules/tournament2/team_details.php

lxp:
PHP Parse error:  syntax error, unexpected '?' in ./modules/tournament2/team_details.php on line 56

Original code before introduction of the syntax error:
https://github.com/lansuite/lansuite/blob/abe8ce608b4a31f8748c0eedd377180641fd973b/modules/tournament2/team_details.php#L50


andygrunwald:
Looks good. Thanks!

PR #88, opened 2017-07-14
Make PHP syntax check fatal

lxp:


lxp:
Okay, syntax check on nightly is failing.
I am going to add an exception for it.

andygrunwald:
This should catch the heavy errors.

lxp:
Yeah, I think so too.
Maybe we should just remove the ext_scripts/gd stuff in the long term, then we might also be able to re-enable fatal checks for nightly.
But I think we should first try to fix code style and make code style checks fatal.

andygrunwald:
Agree here.

PR #92, opened 2017-11-11
Rewrite session handling

hnxfirefly:


andygrunwald:
Thanks @hnxfirefly and sorry for the late reply. I will try to check this PR out in the next days.
Just to let you know that this is still on my road map.

andygrunwald:
This has a dependency to #83 
#83 is also included here.

@hnxfirefly Can you provide a little bit more information and context about this PR here?
Why is it needed to rewrite the session handling? Does the current implementation has any limits? Does the new implementation has benefits? If yes, which one?

hnxfirefly:
The intent behind this is to simplify the session managment into a single one and not have cookies and php-session.  I also have found some bugs in this pr that i have fixed after the merge with #83. 
I already use a version without cookies on www.hnx.at and it works pretty well.
But till i add the latest fixes i'll close this PR.

PR #106, opened 2018-01-05
Upgrade minimum version of MySQL to 5.6.3

andygrunwald:
The system executes the INET6_NTOA() MySQL function.
This function was introduced in MySQL v5.6.3.

See [Changes in MySQL 5.6.3 (2011-10-03, Milestone 6)](https://dev.mysql.com/doc/relnotes/mysql/5.6/en/news-5-6-3.html).

This PR sets MySQL 5.6.3 as a minimum required version.

Error in the UI:
screen shot 2018-01-05 at 21 00 33


M4LuZ:
I would suggest to also include that in the environment check in class_install.php
(That currently checks for PHP version 4.1.2 and reports back version 4.3.0 as required. MySQL module is only checked for existence, but server version is not determined)

andygrunwald:
I added checks in the install tool to check for

* PHP v7.0.0
* MySQL Server v5.6.3

See:
screen shot 2018-01-13 at 21 45 58

@M4LuZ Please check again

PR #107, opened 2018-01-05
Installation: Remove obsolete php performance checks

andygrunwald:
During the installation these PHP settings are checked:

* expose_php
* register_argc_argv
* register_globals
* magic_quotes_gpc

This PR removes all three checks. Why? Because of ...

* [expose_php](https://secure.php.net/manual/en/ini.core.php#ini.expose-php): Exposing the PHP version in a header is nothing bad. It will not consume any more memory and this additional header is not a performance decrease. It is a PHP standard setting. If this would have a negative impact, PHP itself would disable it
* [register_argc_argv](https://secure.php.net/manual/en/ini.core.php#ini.register-argc-argv): This setting is only active when running PHP in a CLI mode. It is not used at all in the web mode (like LanSuite is used). It don't have any kind of relevance to check this during the installation.
* [register_globals](https://secure.php.net/manual/en/ini.core.php#ini.register-globals): This setting was removed in PHP 5.4.0. We require >= PHP 7.0
* [magic_quotes_gpc](https://secure.php.net/manual/en/info.configuration.php#ini.magic-quotes-gpc): This setting was removed in PHP 5.4.0. We require >= PHP 7.0

Removing this code leads to less code maintenance.


M4LuZ:
I would keep the check (and the recommendation to turn it off) for expose_php in. 
Not for performance, but for security reasons.
No benefit in exposing to the internet what version (with potential security flaws) an installation is running.

andygrunwald:
Agree with the point that this is not part of the category Performance.
Even the topic security i would question this, because this don't add a bigger layer of security. Why? Because this is Security by Obscurity.
If an attacker wants to attack your site, it might costs the attacker more time to run all attacks agains the PHP engine, because he don't know the _correct_ version. But this try and error (even automated via metasploit and so on) this only needs more time, but don't add additional security in it. Further more, the attacker is quite quick to discover that we run LanSuite and can check the minimum PHP version which is documented. So there the attacker can check which versions on PHP can be skipped.
But on the counter part, i didn't found a reason why this is activated by default. I [asked via twitter](https://twitter.com/andygrunwald/status/950048338677829639), maybe something comes back. 

So i am fine to change this PR. I merged the master and pushed again. Waiting for the review feedback or merge.

andygrunwald:
@M4LuZ Can you recheck this PR?

andygrunwald:
A screenshot how it looks like

screen shot 2018-01-13 at 21 56 21

M4LuZ:
I do generally agree on the "No security by obscurity part". 
It does not make the system inherently more secure against any determined attacker.
But it does reduce the number of untargeted attacks we can expect.
(It does have advantages to not have your PHP version pop up on shodan or similiar search engines)

PR #108, opened 2018-01-05
Module "stats": Fix wrong mysql_free_result to avoid a sql error

andygrunwald:
The stats module throws an error, because a mysql_free_result is done on a non existing variable.

Screenshot:
screen shot 2018-01-05 at 21 55 32


andygrunwald:
@M4LuZ Agree. free_result is spreaded across the complete codebase and id doesn't hurt here.

PR #110, opened 2018-01-06
PHP Deprecation: Use PHP7 constructors for module boxes, party and stats

andygrunwald:
This PR fixes three PHP 7 deprecation warnings for modules `boxes`, `party` and `stats`:

```
Deprecated</b>:  Methods with the same name as their class will not be constructors in a future version of PHP; menu has a deprecated constructor in ...
```


andygrunwald:
@M4LuZ Can you please recheck this PR?

PR #113, opened 2018-01-06
Reformatted index.php to current coding guidelines and removed obsolete PHP code

andygrunwald:
In this PR i reformatted the index.php according the coding guidelines.
Further more i dropped a lot of PHP code that is not necessary anymore for PHP7.

### DIFF
--- a/index.php
+++ b/index.php
@@ -1,6 +1,12 @@
 <?php
-// Set error_reporting properly
-// Right now we depend on it. This will change in future.
+// Composer autoloading
+require __DIR__ . '/vendor/autoload.php';
+
+// Set error_reporting.
+// It is set to this value on purpose, because otherwise
+// LanSuite might not work properly anymore.
+// Right now we depend on it. This will change in future
+// when the development of LanSuite continues and is getting modernized.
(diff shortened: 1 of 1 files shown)


index.php
```diff
@@ -68,105 +53,88 @@ function myErrorHandler($errno, $errstr, $errfile, $errline)
...
-            'INSERT INTO %prefix%log
-        SET date = NOW(), userid = %int%, type = 3, description = %string%, sort_tag = "PHP-Fehler"',
-            (int)$auth['userid'],
+        $db->qry('
+            INSERT INTO %prefix%log
+            SET date = NOW(),
+                userid = %int%,
+                type = 3,
+                description = %string%,
+                sort_tag = "PHP-Fehler"',
```
M4LuZ: Should we change this from "PHP-Fehler" to "PHP-Error" or maybe a general "PHP"?
Did a quick search and this is the only place in the source code where it is used.

Field is only used in the Log viewer for sorting purposes and the selection list is auto-filled based on unique entries, so no separate edit needed there..

andygrunwald: True. This needs to be investigated further what it is used to be. For this i created #115

PR #119, opened 2018-01-13
Docker: Install PHP extension "xdebug" in our image

andygrunwald:
[xDebug](https://xdebug.org/) is one of the most used debugger extensions for PHP.
With this error messages are shown with a proper stacktrace. In this way it is way easier to determine the root cause of e.g. an SQL free result error.

In the readme it is stated that the Docker container is not used for production right now.
In the long run we can create two dockerfiles: One for production, one for development.
For now we are in a phase to get it up and running again on newer versions of PHP and MySQL.
So for people who are testing LanSuite and reporting issues, this might be a better thing, because then they are able to provide a stack trace.


M4LuZ:
Definitive agreement on this.
Buuut:
Won't the remote host functionality need additional changes of the container network settings (or parameters to set the remote host correctly)?
See https://xdebug.org/docs/remote#remote_host

andygrunwald:
I will have a look. https://gist.github.com/chadrien/c90927ec2d160ffea9c4 might be useful here.
I will test it and readjust it here.

andygrunwald:
@M4LuZ I added `remote_enable` and `remote_connect_back` to the settings.
I configured PHPStorm and enabled debugging with breakpoints.

Works as expected. If you want to configure it as well, check out [Debugging within a PHP Docker Container using IDEA/PhpStorm and Xdebug](https://blog.philipphauer.de/debug-php-docker-container-idea-phpstorm/).

Please review again.

Dockerfile
```diff
@@ -12,31 +12,35 @@ RUN apt-get update \
     && docker-php-ext-install -j$(nproc) mysqli snmp \
     && docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ \
     && docker-php-ext-install -j$(nproc) gd \
+    # Development extensions
+    && pecl install xdebug-2.5.5 \
```
M4LuZ: Is the reference to a particular xdebug version intentional? 
It does prevent problems when there is an incompatible update, but could cause the same if any of the other unversioned references receives an incompatible update

andygrunwald: As far as I know, this is the standard way of installing xDebug. In pecl (https://pecl.php.net/package/xdebug) there is no "latest" for the binaries.
Furthermore, with this, we can run controlled upgrades and upgrades need a PR as well.

M4LuZ: According to https://xdebug.org/docs/install there is `pecl install xdebug`.
We could either leave it at the most current version (and then fix issues with a PR when or if they ever appear) or fix all of the packages and then take care about updates manually. As this requires a certain discipline and regularity, I would go for number one, given that those packages should be fairly stable and to save maintenance effort

M4LuZ:
Ah, I would also suggest to change the README.md from suggesting that the docker image *should* not be used in production to a definitive warning that it **must** not be used in production, as having xdebug up and running (among other settings) allows easy remote exploitation

andygrunwald:
@M4LuZ I fixed the Xdebug version and rewrote the warning in the Readme.
I still stuck to "should not", because I will not forbid someone to do this. We only can make the Warning clear. What people do is still their choice. This is what I encounter as best practice from other open source documentation also (mostly in suggestions for database backends and so on).

M4LuZ:
Of course we can't prohibit the usage, but we should make abundantly clear that it is a very bad idea to do so. And I feel that "should" is a wording too soft for that. 
But not really worth a discussion ;)

andygrunwald:
Thanks for the review @M4LuZ

PR #100, opened 2017-12-17
Added phpdoc into composer.json and streamlined application requirements

andygrunwald:
What we did here:

* Added `phpdoc` to the development requirements into our composer / package manager
* Added PHP >=7.0 + extensions gd, mysqli and snmp
* Added `bin` folder for scripts installed by external dependencies

PR #121, opened 2018-01-13
Composer: Added application requirements

andygrunwald:
What we did here:

* Added PHP =7.0.* + extensions gd, mysqli and snmp
* Added `bin` folder for scripts installed by external dependencies

This was part of https://github.com/lansuite/lansuite/pull/100 before.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,10 +12,11 @@
         }
     ],
     "require": {
-        "php-64bit": "^7.0",
+        "php": "^7.0",
         "ext-gd": "*",
         "ext-mysqli": "*",
         "ext-snmp": "*",
+
         "smarty/smarty": "^3.1",
         "geshi/geshi": "^1.0.9"
     },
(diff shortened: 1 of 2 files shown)


andygrunwald:
@M4LuZ This only sets the necessary requirements. Most (not all) requirements are checked during installation anyway. So it is just in line with defining them here as well.
This can also act as a self protection, because if a package don't align with the requirements here, we are not able to install / require it. With this bugs can be detected earlier.

composer.json
```diff
@@ -12,7 +12,15 @@
         }
     ],
     "require": {
+        "php-64bit": "^7.0",
```
M4LuZ: While I do think that most to all systems should be 64bit by now, is this really a requirement at this point?

andygrunwald:
@M4LuZ Good point. I updated the PR. Please review again.

PR #122, opened 2018-01-13
Git: Added a few comments into gitignore

andygrunwald:


andygrunwald:
@M4LuZ This adds only a few comments into the .gitignore file. No change in functionality. Just to provide more context for other engineers.

M4LuZ:
That really needed no explanation, but thanks nevertheless ;)

PR #125, opened 2018-01-13
Installation: Fix MySQL error while checking for super admin if the lansuite installation is not completed yet

andygrunwald:


andygrunwald:
Okay, after a fresh and clean installation, this blocks step 7 of the installation.

PR #126, opened 2018-01-13
Fix #117: Query to stats_auth database table incompatible with sql_mode=only_full_group_by

### LINKED ISSUE #117: incompatible with sql_mode=only_full_group_by
Clean install via git clone (commit b51cbb9)
MySQL v5.7.20-0ubuntu0.16.04.1

Fehler: 
SQL-Failure. Database respondet: Expression #2 of SELECT list is not in GROUP BY clause and contains nonaggregated column '<db>.ls_stats_auth.lasthit' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by

Query: 
SELECT userid,lasthit FROM ls_stats_auth WHERE login = "1" AND (lasthit > 1515528332 OR lastajaxhit > 1515528872) AND userid > 0 GROUP BY userid

andygrunwald:
See #117

### DIFF
--- a/inc/classes/class_auth.php
+++ b/inc/classes/class_auth.php
@@ -52,9 +52,18 @@ class auth
 
         // Better handle it here, otherwise its an DB-Query for each $dsp->FetchUserIcon()
         $res = $db->qry(
-            'SELECT userid,lasthit FROM %prefix%stats_auth
-            WHERE login = "1" AND (lasthit > %int% OR lastajaxhit > %int%) AND userid > 0
-            GROUP BY userid',
+            'SELECT
+                userid,
+                lasthit
+            FROM %prefix%stats_auth
+            WHERE
+                login = "1"
+                AND (
+                    lasthit > %int%
+                    OR lastajaxhit > %int%
+                )
+                AND userid > 0
+            GROUP BY userid, lasthit',
             $this->timestamp - 60*10,
             $this->timestamp - 60*1
         );


inc/classes/class_auth.php
```diff
@@ -52,9 +52,18 @@ public function __construct($frmwrkmode = "")
...
+                lasthit
+            FROM %prefix%stats_auth
+            WHERE
+                login = "1"
+                AND (
+                    lasthit > %int%
+                    OR lastajaxhit > %int%
+                )
+                AND userid > 0
+            GROUP BY userid, lasthit',
```
M4LuZ: I suspect this does not return the same user count if a user is logged in in two separate sessions. In that case  both fields in the GROUP BY clause are unique (unless last access was done in the same second in both sessions) returning two results instead of one as previously

andygrunwald: I will re-check this in the next days and will find a good solution for this.

M4LuZ: I guess either a direct selection by the MySQL time functions (something like `SELECT DISTINCT(userid) WHERE TIMESTAMPDIFF(MINUTE,lasthit,NOW())<10 AND TIMESTAMPDIFF(MINUTE,lastajaxhit,NOW())<1)`(so that lasthit is not needed in the later code) or a direct counting in MySQL and removal of the later PHP code should do the trick. 
But haven't tested it and this by far from being a critical issue, of course.
I would rather suggest to raise a ticket and spend time on it whenever we got really nothing better to do ;)

andygrunwald: I fixed the regression in https://github.com/lansuite/lansuite/pull/139

PR #129, opened 2018-01-19
Install: Fix design dropdown in installation

andygrunwald:
During the installation, the design dropdown is empty.
This PR fix the issue.

Before:
screen shot 2018-01-19 at 20 14 28

After:
screen shot 2018-01-19 at 20 13 10


M4LuZ:
Just for my understanding: 
Am I guessing correctly that the problem was the usage of a forward slash as directory separator instead of the OS-independent DIRECTORY_SEPARATOR causing issues when running on Windows?
And all the other modifications are only cosmetic?

andygrunwald:
@M4LuZ The main change / bug was:

```
is_dir($currentDesign)
```
vs.
```
is_dir($designPath . $currentDesign)
```

The code was checking if there was a directory called "simple" (for example). But the directory is located in `design/simple`.
The other changes are just cosmetic.

PR #131, opened 2018-01-21
Improve Docker setup

mre:
Corrections to the Docker setup

* Fix markdown formatting for `<Your-Docker-IP>` in `README`.
* Install composer in lansuite Docker image.
* Clean-up apt resources after installation in Dockerfile to reduce image size.
* Set workdir to `/code` in lansuite image to ease usage of commands like `docker-compose run php bash` and `docker-compose run php composer install`.
* `COPY` application files into lansuite container to make it a self-hosted image (no runtime volume mounts necessary for simply using it).
* Add a build context to the `docker-compose.yml`. This way, new versions of the `lansuite/lansuite:latest` images will be built automatically with `docker-compose build` and in case Dockerhub is not reachable.
* Remove network declarations from `docker-compose.yml`, as they were the default.
* Add `composer install` step to installation instructions. Although we run `composer install` at image build time, this is still necessary, since we overwrite the `/code` volume with our local code at runtime for development.


M4LuZ:
Hi @mre,

first of all thanks a lot for your participation in this project!

Merge request looks fine for me except for one part:
`COPY application files into lansuite container to make it a self-hosted image (no runtime volume mounts necessary for simply using it).`
I think this was intentional.
As the docker image is targeted as easy development environment the mount is the easiest way to have modifying access to the source code hosted in the container from an IDE. Having the files inside the container would require to have an additional service (SMB, FTP, whatever) running to make the source code available for editing.

andygrunwald:
Thanks @mre for the PR. And thanks @M4LuZ for the review.

@M4LuZ To provide you some context:  @mre is a friend of me. @mre, @timglabisch and i had a "hacking session" on saturday where i showed them the LanSuite product.

Related to the `COPY` cmd. I had a chat with @mre exactly about this. He said if you want to use the "development version" and change the code, that you can overwrite the `/code` dir with a mount. And i think this is quite nice, because then

* this image can be use for every end user to test / play / set it up very easyly
* developers can use this image for developing

This needs to build in into this PR, but i or @mre will test it. What do you think about this @M4LuZ ?

Dockerfile
```diff
@@ -9,4 +11,29 @@ RUN apt-get update \
         snmp \
     && docker-php-ext-install -j$(nproc) mysqli snmp \
     && docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ \
-    && docker-php-ext-install -j$(nproc) gd
\ No newline at end of file
+    && docker-php-ext-install -j$(nproc) gd \
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-get clean
+
+RUN docker-php-ext-install zip
```
andygrunwald: A comment that the Composer setup starts here would be nice.
Because the `zip` extension is required by Composer, but not by LanSuite.

Dockerfile
```diff
@@ -9,4 +11,29 @@ RUN apt-get update \
...
+
+ENV COMPOSER_ALLOW_SUPERUSER 1
+ENV COMPOSER_HOME /tmp
+ENV COMPOSER_VERSION 1.6.2
+
+RUN curl -s -f -L -o /tmp/installer.php https://raw.githubusercontent.com/composer/getcomposer.org/b107d959a5924af895807021fcef4ffec5a76aa9/web/installer \
+ && php -r " \
+    \$signature = '544e09ee996cdf60ece3804abc52599c22b1f40f4323403c44d44fdfdd586475ca9813a858088ffbc1f233e9b180f061'; \
```
andygrunwald: Wouldn't it make sense to put the signature into an Env var as well?
Does the signature relates to the specific Composer version or to the installer?
So is the signature changing, when installing a newer version of composer?

README.md
```diff
@@ -38,12 +38,13 @@ $ # Add the content of the example configuration file below into ./inc/base/conf
 $ chmod 0777 ./inc/base/config.php
 $ chmod -R 0777 ./ext_inc/
 $ docker-compose up
+$ docker-compose run php composer install
```
andygrunwald: Isn't this triggered by the Dockerfile already?

M4LuZ:
> What do you think about this @M4LuZ ?

I do see the advantage to make this an evaluation platform. We can rather ask someone interested in development to do additional setup steps than someone that wants to do an evaluation with as little work as possible.

That is of course ignoring the elephant in the room that this doesn't work on Windows .At all.
A readymade bundle with/for XAMPP (or alternatives) would be perfect for that.
But that is a different story and not related to this PR ;)

mre:
Thanks for your input @M4LuZ.
We can have both: a "production" image with the code being part of the final image and a development image, which reflects code changes in realtime.
This is done by having a `COPY` inside the Dockerfile and a volume pointing to the same location within the container for development. The volume will overwrite the copy. It's a win-win. 😉 
For Windows, it should work the same without modifications.

Dockerfile
mre: I don't know. That's probably something to discuss at https://github.com/composer/docker/, which is where I took that part from. 😄

README.md
mre: Yes, but depending on when the upstream image was last pushed, that might be outdated by now.
Also, when the `composer.json` gets modified, a rerun might also be needed.
Therefore, to avoid unexpected side-effects it's better to add it to the docs, in my opinion.

M4LuZ:
I just noticed docker for windows is existing and available.
What sorcery is this? 😲 
Then I'm fine with whatever you come up here.

> We can have both: a "production" image with the code being part of the final image and a development image, which reflects code changes in realtime.

Fine with that, just call it "evaluation" instead of production, otherwise some *will* use it productively...

andygrunwald:
Thanks for this contribution @mre !
I have a couple of small other ideas for the docker setup. I will discuss those today with you in person and we make a PR.

PR #135, opened 2018-01-24
Remove function htmlspecchars() and moved d() to inc/Functions/Debug.php

andygrunwald:
Those functions are not called / used in the complete codebase.


byte666:
Langsam. d() ist die Zentrale Debugfunktion. Warum wird sowas gelöscht?

BenniGotchi:
the d() function is not called in the checked in version, but it's a tool that can be used while developing new functionallity / tracing bugs, and so on...
It was very handy in the past, because you can easily add debugging output to LS...

andygrunwald:
Modern debugging practices within PHP  are offered with things like xDebug.
xDebug shows you complete stack traces and have an instep debugger which can be configured in nearly every IDE. See #119

M4LuZ:
> Modern debugging practices within PHP are offered with things like xDebug.

True, but I do see the advantage of having an easy way to drop in additional logging in case you hit an issue on the productive system (where xDebug and such is normally not present).
There are certainly some nice frameworks, but embedding them in the code doesn't get much simpler than wrapping `d()` around your code

I am not a big fan of removing stuff that is not broken and does serve a purpose.
Thus I would vote to keep it in unless we have a (better) replacement

andygrunwald:
You have a point. Then I suggest moving the `d()` function to another file and include this in an `index.php` or move the call to a static class that can be autoloaded. Because functions can't be autoloaded (as far as I know, need to look it up). (see #137).

What about the function `htmlspecchars`. What is your opinion about this?

andygrunwald:
New update.
What is this PR doing and why:

* Removed function `htmlspecchars()`: This is not called in the codebase.
* Moved function `d()` to inc/Functions/Debug.php: We move the debug function d() into an own file and away from the class_debug.php, because every class should be located in an own file. To enable proper autoloading later. This is also the reason why the folder `Functions` and the File `Debug.php` is written uppercase. This follows the PHP standards PSR-0 and PSR-4.

@BenniGotchi @byte666 Whats your opinion here?

andygrunwald:
Any update here on a review? /cc @byte666 @M4LuZ @BenniGotchi

M4LuZ:
I thought I left a comment here, time to redeem that...

As htmlspecchars() is not used and escaping in general is being discussed (and most probably reworked) it is fine to be removed.
code for debug functionality is just being moved, thus no change in functionality, thus not much to complain about here

byte666:
M4LuZ already mentioned the d() funktion is important for systems without xDebug (special productive). This is why we created this function. There are some features you can just use if you edit the file. (write2file if you get a white Screen, etc.) Not nice, but sometimes you will need this. Place d() in a separate file... no Problem. But such global functions should be rare and well documented.

> his is also the reason why the folder Functions and the File Debug.php is written uppercase. This follows the PHP standards PSR-0 and PSR-4.

Really? Sounds unworldly. I learnd in the last 25 years... in filesystems ALWAYS just use lowercase, no blanks, no special chars. Over time you would get always in trouble, special on mixed environments. Windows/Unix, coders make mistakes, etc. Failtures based on this maybe will not happen often, but can be really stubborn. :)

M4LuZ:
> Really? Sounds unworldly. I learnd in the last 25 years... in filesystems ALWAYS just use lowercase, no blanks, no special chars.

Yes, that is the easiest way to avoid these issues completely and I thought that would also be reflected in the coding standards.
But interestingly [PSR-1 ](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-1-basic-coding-standard.md) does indeed declare that `Class names MUST be declared in StudlyCaps.`
And [PSR-4](https://www.php-fig.org/psr/psr-4/) says that `All class names MUST be referenced in a case-sensitive fashion.` In other words: Use uppercase characters for both class and filenames...

Following that guide shouldn't harm us that much with Linux being both the primary development and (continuous) integration platform, as these problems are bound to appear before being merged here. 
And Windows just doesn't care (tm) 😉

andygrunwald:
_Many_ open source systems follow these standards (e.g., Symfony2, Flow, TYPO3, Laravel and more). Those systems are way more in use than LanSuite. And they run without issues here.

I resolved the conflict here. If no one disagrees, this one is ready to merge.

M4LuZ:
> Those systems are way more in use than LanSuite

THAT IS BLASPHEMY! 😉 

And to the topic at hand: Fine for me ;)

andygrunwald:
Good to merge here? @byte666 @M4LuZ ?

andygrunwald:
I will continue with a merge here after an approval by @M4LuZ

PR #136, opened 2018-01-24
Removed unused constants

andygrunwald:
Those constants are not called / used in the complete codebase.


andygrunwald:
Any update here on a review? /cc @byte666 @M4LuZ

JochenJung:
I think OPTIONAL and SELECTED where intended for masterform to be passed to the method which adds an input row, defining whether that item is Optional / Selected or not.

But as its not used anymore: Sure, delete it.

The other constants where not introduced by me, I think. So I do not know about them.

M4LuZ:
Having the option to define basic properties for Input rows seems to be a good option to implement.
You say that these are not in use anymore. 
But aren't they used as another option was implemented or because the required functionality has never been implemented?
Because in the later case I would rather put them to use instead of removing them

JochenJung:
I'm not sure. You would need to check in class_masterform.

andygrunwald:
> But aren't they used as another option was implemented or because the required functionality has never been implemented?
> Because in the latter case I would rather put them to use instead of removing them

Even is the latter one is the case, I would go _now_ for removing and once someone is using them or is reworking the master form usage, then introduce them with proper usage in the codebase.
This acts now as a code cleanup and is able to focus on the code that is in use.

andygrunwald:
Any update here? Ready for a merge? @M4LuZ ?

M4LuZ:
> M4LuZ approved these changes 18 days ago 

'Nuff said ;)

andygrunwald:
Just a final confirmation, because a discussion started after this.

Issue #132, opened 2018-01-22
Introduce autoloading of Classes with Composer


andygrunwald:
Autoloading is a well known and well used feature inside PHP.
See https://secure.php.net/manual/en/language.oop5.autoload.php for details.

## Expected Behavior

No `require_once` or include calls more in the code base.
Autoloading of functions and classes in the codebase.

## Current Behavior

Every class and function is required by an include call.

## Possible Solution

Load classes and functions with the help of composer.
See http://phpenthusiast.com/blog/how-to-autoload-with-composer

andygrunwald:
This is on the road and will be done during the refactorings.
I will close this ticket. If someone wants to raise their voice, feel free and reopen :)

PR #137, opened 2018-01-24
Added class autoloading for inc/classes folder via composer

andygrunwald:
All classes from `inc/classes` are now autoloaded via composer. See https://github.com/lansuite/lansuite/issues/132 for details.

This is the first PR out of a few. I plan to make multiple PRs to keep the changeset smaller and more reviewable.

Further more the manual approach to add the classes to composer is only the first step.
End goal is to have all classes autoloaded via PSR4 with name spaces.


andygrunwald:
Any update here on a review? /cc @byte666 @M4LuZ

PR #139, opened 2018-01-28
Fix #117: Query to stats_auth database table incompatible with sql_mode=only_full_group_by

### LINKED ISSUE #117: incompatible with sql_mode=only_full_group_by

andygrunwald:
This PR fixed the comments by @M4LuZ in #126.

If the same user is logged in multiple times with different sessions, it would return a wrong count.
This query now works on

* MySQL 5.6
* MySQL 5.7 (incl. sql_mode=only_full_group_by)

### DIFF
--- a/inc/classes/class_auth.php
+++ b/inc/classes/class_auth.php
@@ -12,14 +12,8 @@ require_once("inc/classes/class.crypt.php");
  * @access public
  * @todo Change uniqkey from md5(password) to an extra Field
  */
-class auth
-{
+class auth {
 
-  /**#@+
-   * Intern Variables
-   * @access private
-   * @var mixed
-   */
     public $auth = array();                 // Userdaten im Array
     public $timestamp;                      // Zeit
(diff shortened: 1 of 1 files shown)


andygrunwald:
True. I was not able to fix this in a different way, except disabling `only_full_group_by`. But for now, it is returning the right result again.

PR #147, opened 2018-02-23
Extract class barcode_system to an own file and add into autoloading

andygrunwald:
This PR extracts the PHP class `barcode_system` into an own file and adds this class into autoloading.
Furthermore, the constant `BARCODE_PATH` was moved to the PDF module, because this is the only location where the constant is used.

PSR-1 say:
> This means each class is in a file by itself, and is in a namespace of at least one level: a top-level vendor name.

Source: https://www.php-fig.org/psr/psr-1/

Namespaces are not introduces (yet), but this is one more commit into this direction.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
             "inc/classes/class.crypt.php",
             "inc/classes/class_auth.php",
             "inc/classes/class_barcode.php",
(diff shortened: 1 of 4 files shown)


andygrunwald:
@M4LuZ Any updates here?

PR #153, opened 2018-03-17
Rework of XML class: Reformatting, PHPDoc for methods and unit tests

andygrunwald:
This PR contains no functionality change.

- No functionality change!
- Wrote a couple of unit tests


andygrunwald:

PR #154, opened 2018-03-17
Fix #148: Don't enforce php-snmp, only suggest it.

### LINKED ISSUE #148: Should presence of php-snmp be enforced by composer?
## Expected Behavior

As the PHP module SNMP is only required for the network module (which has questionable functionality at the moment) I would suggest that presence is not enforced by composer.

## Current Behavior

`composer up` stops when PHP SNMP module is not present and activated

## Possible Solution
As presence is also checked during initial setup (as optional component) it is not required to enforce this in composer also. Sadly there is not (yet) an option to define requirements as optional in composer, so the decision would be in or out...

run `composer up` after checkout without having the SNMP module active

andygrunwald:
See #148 for details.

The PHP extension `snmp` is only used for the Network Operation Center module.
If the extension is not present, composer will suggest the extension.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,6 @@
         "php": "^7.0",
         "ext-gd": "*",
         "ext-mysqli": "*",
-        "ext-snmp": "*",
 
         "smarty/smarty": "^3.1",
         "geshi/geshi": "^1.0.9"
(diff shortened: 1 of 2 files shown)


andygrunwald:
This needs to be merged before #153

M4LuZ:
> This needs to be merged before #153

Well...shit! 
I started at the bottom of the queue and according to #153 it already includes this?

andygrunwald:
@M4LuZ Everything is fine :)
The change is already in. See https://github.com/lansuite/lansuite/blob/master/composer.json#L31
No need for this PR anymore.

You can now go ahead with the other PRs :)

PR #155, opened 2018-03-17
Rework of Translation class: Reformatting, PHPDoc for methods, Cleanup and unit tests

andygrunwald:
- No functionality change in methods that stayed
- Removed method "get_calling_module", because it was not in use
- Removed method "xml_var_merge", because it was not in use
- Removed method "get_trans_file", because it was not in use
- Added a couple of unit tests
- Extracted functions t() and t_no_html() into an own file (=> inc/Functions/T.php)


andygrunwald:

M4LuZ:
Fine with this as:
* No functional change
* `get_calling_module` is just a wrapper for debug_backtrace and marked as NOT WORKING
* `xml_var_merge` is not used and I think also at the wrong place (translation related?)
* `get_trans_file`is not used and the idea of search with a hashed key is implemented already in t()

PR #156, opened 2018-03-17
Rework of Sec class: Reformatting, PHPDoc for methods, Cleanup and unit test

andygrunwald:
- Made check_blacklist() testable by moving the die() to the caller
- Added one unit tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 # LanSuite configuration
 /inc/base/config.php
 
+# Installed by vendors
+bin/phpunit
+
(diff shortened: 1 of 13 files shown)


andygrunwald:

M4LuZ:
Raised #162 for this

andygrunwald:
> But there is no call to add an IP to the blacklist, so unless done by hand, this doesn't actually do anything.
> At least that is what I was able to see so far.

I agree so far.
The config can be used to add IP addresses (manually).

PR #157, opened 2018-03-17
Rework of Plugin class: Reformatting, PHPDoc for methods, fixed casting bug

andygrunwald:
- No functionality change in methods that stayed
- Fixed on (int) casting bug
Old:
```
if ($index == -1) {
    (int)$index = $this->currentIndex;
}
```

New:
```
if ($index == -1) {
    $index = (int) $this->currentIndex;
}
```

### DIFF
--- a/inc/classes/class_plugin.php
+++ b/inc/classes/class_plugin.php
@@ -1,28 +1,41 @@
 <?php
 
 /**
- * Class PLUGIN for Lansuite
+ * Class plugin
  *
- * This Class generates the needed Dataarray to load the Pluginfiles
- *
- * @package lansuite_core
- * @author bytekilla
- * @version $Id$
- * @access public
+ * This class generates the needed data array to load the plugin files.
  */
 class plugin
 {
 
-  /**
-   * Intern Variables
-   * @access private
-   * @var mixed
-   */
-    public $modules    =     array();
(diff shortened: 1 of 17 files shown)


andygrunwald:

inc/classes/class_plugin.php
```diff
@@ -1,28 +1,41 @@
 <?php
 
 /**
- * Class PLUGIN for Lansuite
+ * Class plugin
  *
- * This Class generates the needed Dataarray to load the Pluginfiles
- *
- * @package lansuite_core
- * @author bytekilla
```
M4LuZ: Is there a good reason to remove code attribution to @byte666?

andygrunwald: In my opinion the @author-tags are better filled by a `git blame` or a `git log`.
The history of LanSuite is part of this repository.
The software is developed by a team and it is a team effort.
If i want to know who wrote it, i will fire a `git blame`.
I am thankful for every line that was written by someone here.

The sentene `This Class generates the needed Dataarray to load the Pluginfiles` was removed, because i have to admit: For me it says nothing useful. What Dataarray?
The context is missing IMO.

M4LuZ: OK, fine by me ;)

PR #161, opened 2018-03-18
Fix #149: Add checks to avoid white screen on install

### LINKED ISSUE #149: install.php only presents WSOD in many cases
From my own experience most of the issues that cause the start of the installation script to fail are not handled gracefully.
Most of the time a blank screen is presented to the user without any reference what went wrong.
(I also just have a case where also no error messages end up in the php.log, making this even nicer...)

## Expected Behavior

The user should be able to identify what went wrong easily

## Current Behavior

Installer ends up with a white page when:
* smarty template directory is not writable (`./ext_inc/templates_c`)
* composer dependencies have not been added (and thus inclusion of the autoloader fails)
* the problem above I have not yet identified.

## Possible Solution

Either errors occuring before the install environment should be handled gracefully or additional environment checks should be implemented before trying to load the installation.
Or - as the easy workaround -  we just make sure that the PHP-Error level is high enough and error messages are shown on the client.

try to install lansuite in a not completely prepared environment ;)

## Your Environment

* Version used: Git/MASTER
* Operating System and version: Fedora 27
* Enabled features:

M4LuZ:
Added some small preliminary checks to install.php to avoid issues that can cause it to end up just showing a blank screen

fixes #149

### DIFF
--- a/install.php
+++ b/install.php
@@ -1,17 +1,21 @@
 <?php
 
-//raise error reporting (everything could be interesting...
+//raise error reporting - could be interesting and this is autmomatically reset when index.php is executed
 error_reporting(E_ALL);
 
-//check some known issues first to make sure that we can execute index.php...
 
+/**
+* Check for common failures
+*
+* Small functtion to check against a number of dependencies that cause install.php to fail without showing any error details
+* @return bool check result. True if error found, false if all OK
+*/
 function checkdeps(){
 
 	$error =false;
 	$errmsg='The following issues were found while preparing to run the installation:<br/>'. PHP_EOL;
 
 	//check if composer was executed (./vendors existing)
-
 	if (!is_dir(__DIR__ . '/vendor')) {
 
 		$error = true;
@@ -25,12 +29,11 @@ function checkdeps(){
 	}
 
 	//check error state and return stored messges if problem found
-
 	if ($error) echo $errmsg;
 	return $error;
 }
 
-
+//execute install.php only if dependencies are met
 if (!checkdeps()){
 	$_GET["mod"] = "install";
 	include_once("index.php");
--- a/install.php
+++ b/install.php
@@ -1,3 +1,37 @@
 <?php
-$_GET["mod"] = "install";
-include_once("index.php");
+
+//raise error reporting (everything could be interesting...
+error_reporting(E_ALL);
+
+//check some known issues first to make sure that we can execute index.php...
+
+function checkdeps(){
+
+	$error =false;
+	$errmsg='The following issues were found while preparing to run the installation:<br/>'. PHP_EOL;
+
+	//check if composer was executed (./vendors existing)
+
+	if (!is_dir(__DIR__ . '/vendor')) {
+
+		$error = true;
+		$errmsg .= __DIR__ . '/vendor directory is missing. Did you run composer?<br/>' .PHP_EOL;
+	}
+
+	//check if smarty-directory is writable
+	if (!is_writable(__DIR__ . '/ext_inc/templates_c')) {
+		$error = true;
+		$errmsg .= __DIR__ . '/ext_inc/templates_c seems not to be writable by the PHP user. Please confirm that file and folder access rights are applied correctly<br/>' . PHP_EOL;
+	}
+
+	//check error state and return stored messges if problem found
+
+	if ($error) echo $errmsg;
+	return $error;
+}
+
+
+if (!checkdeps()){
+	$_GET["mod"] = "install";
+	include_once("index.php");
+}
--- a/install.php
+++ b/install.php
@@ -7,7 +7,7 @@ error_reporting(E_ALL);
 /**
 * Check for common failures
 *
-* Small functtion to check against a number of dependencies that cause install.php to fail without showing any error details
+* Small function to check against a number of dependencies that cause install.php to fail without showing any error details
 * @return bool check result. True if error found, false if all OK
 */
 function checkdeps(){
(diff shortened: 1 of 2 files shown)


install.php
```diff
@@ -1,3 +1,40 @@
 <?php
-$_GET["mod"] = "install";
-include_once("index.php");
+
+//raise error reporting - could be interesting and this is autmomatically reset when index.php is executed
```
andygrunwald: I would favor comments like `// Raise error ...`.
The reason is readability. It is like a sentence in an article or in a book. There you expect also a "Whitespace" in front of the sentence (like `This is super. And here it goes ...`. And not `This is super.And here it goes ...`. What do you think about this?

Furthermore, `autmomatically` => `automatically`

install.php
```diff
@@ -1,3 +1,40 @@
...
+error_reporting(E_ALL);
+
+
+/**
+* Check for common failures
+*
+* Small function to check against a number of dependencies that cause install.php to fail without showing any error details
+* @return bool check result. True if error found, false if all OK
+*/
+function checkdeps(){
```
andygrunwald: I would suggest a more declarative name like `checkInstallationDependencies` or `checkDependencies`.

And wouldn't it make sense to place this function in an own file inside `inc/Functions/`?
With this we have a clear split. We just do a `require_once(...)` here.

install.php
```diff
@@ -1,3 +1,40 @@
...
+
+//raise error reporting - could be interesting and this is autmomatically reset when index.php is executed
+error_reporting(E_ALL);
+
+
+/**
+* Check for common failures
+*
+* Small function to check against a number of dependencies that cause install.php to fail without showing any error details
+* @return bool check result. True if error found, false if all OK
```
andygrunwald: Afaik PHPDoc says, that we insert a blank line between the descriptions and the parameters/return values.

install.php
```diff
@@ -1,3 +1,40 @@
...
+*/
+function checkdeps(){
+
+	$error =false;
+	$errmsg='The following issues were found while preparing to run the installation:<br/>'. PHP_EOL;
```
andygrunwald: The `$var=value` is inconsistent.
Sometimes `$var=value` or `$var =value` or `$var = value`. Personally i think `$var = value` is more readable. Do you agree?

install.php
andygrunwald: Oh. This applies to all `//` comments in this PR.

install.php
```diff
@@ -1,3 +1,40 @@
...
+
+	//check if composer was executed (folder ./vendor created)
+	if (!is_dir(__DIR__ . '/vendor')) {
+
+		$error = true;
+		$errmsg .= __DIR__ . '/vendor directory is missing. Did you run composer?<br/>' .PHP_EOL;
```
andygrunwald: Why we put a `<br />` at the end?

install.php
```diff
@@ -1,3 +1,40 @@
...
+	}
+
+	//check if smarty-directory is writable
+	if (!is_writable(__DIR__ . '/ext_inc/templates_c')) {
+		$error = true;
+		$errmsg .= __DIR__ . '/ext_inc/templates_c seems not to be writable by the PHP user. Please confirm that file and folder access rights are applied correctly<br/>' . PHP_EOL;
+	}
+
+	//check error state and return stored messages if problem found
+	if ($error) echo $errmsg;
```
andygrunwald: I suggest using curly braces (`{`) even in one line if statements.
The reason is to avoid bugs, once we extend this clause. A quick change then can introduce bugs, because only the first statement is executed in such a condition if no curly braces are there.

Next to this:
What do you think about return an empty string, once no error occurred and a string with the error message, if an error is there? Instead of `echo` the message here?
Then we could a) test this function (e.g. with a unit test) and b) make a proper output later on (with templates or anything).
The alternative would be to throw an Exception. Just an idea. Let me know what do you think.

M4LuZ:
Thanks for the review!
I already felt that I missed some of our own requiremenst, but oh boy....;)
Will have a look and fix accordingly

install.php
M4LuZ: That one is fairly simple to answer:
As there is no previous/later page output, this is shown as-is.
The browser irgnores the EOL but <br/> does still enforce a line break, so that the different text pieces are shown in their own line, now matter how viewed

andygrunwald: Thanks! Fine for me :)

install.php
M4LuZ: I was in doubt if I should even write a function. 
Not like the code is repeatly called or could be used elsewhere.
But thinking about it now :
This issues could re-appear during upgrade scenarios (e.g. unpack new release as root and forget to chown afterwards).
In that case it would fit more into index.php. 
But on the other hand I don't really want to run that check on every execution.
So we could also put it into the install class as static function (to avoid creating an object with its dependencies) and provide an full environment checker (that also runs the other install checks) instead.
Your opinion?

install.php
M4LuZ: yup ;)

M4LuZ:
OK, changed stuff accordingly.
Would be happy for a review.
Will give the changed files another round of tests then before finally merging it

andygrunwald:
Thank you :)

M4LuZ:
Ahem....
> Will give the changed files another round of tests then before finally merging it

**Means that this is at the moment untested code.** 
My test system is a few hundred kilometers away so I was not able to go through any test after refactoring :/

andygrunwald:
I can run a test tonight on this. Will report the results here :)
Otherwise we will revert it. No worries :)

M4LuZ <notifications@github.com> schrieb am Do. 22. März 2018 um 16:13:

andygrunwald:
@M4LuZ Indeed. There was a small bug in it. `__DIR__`refers to the directory where the executed file is located. I fixed it in https://github.com/lansuite/lansuite/pull/169 and included screenshots in it.

PR #163, opened 2018-03-21
Rework of masterdelete class: Reformatting, PHPDoc for methods, removed dead code

andygrunwald:
- Removed dead code

### DIFF
--- a/inc/classes/class_masterdelete.php
+++ b/inc/classes/class_masterdelete.php
@@ -2,45 +2,67 @@
 
 class masterdelete
 {
-    public $References = array();
-    public $SubReferences = array();
-    public $DeleteIfEmpty = array();
+    /**
+     * @var array
+     */
+    public $References = [];
+
+    /**
+     * @var array
+     */
+    public $SubReferences = [];
+
+    /**
+     * @var array
+     */
+    public $DeleteIfEmpty = [];
+
+    /**
+     * @var int
+     */
     public $LogID = 0;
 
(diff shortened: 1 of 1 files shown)


inc/classes/class_masterdelete.php
```diff
@@ -125,28 +158,40 @@ public function Delete($table, $idname, $id)
...
+
             unset($_SESSION['md_referrer']);
+
             return $res;
         }
     }
 
+    /**
+     *
+     * TODO Question for ASK_DELETE AND ASK_SET0
```
M4LuZ: there is AFAIK also `@ToDo`f for that

andygrunwald: Yes, many IDEs support both versions: `@todo` or `TODO`.
Afaik there is no strict convention over it.
I mostly search for "todo" which catches both.
I don't have a preference. For me it is just important, to mark this as a task.

M4LuZ: I could only find `@todo` the [PHPdoc documentation](http://docs.phpdoc.org/references/phpdoc/tags/todo.html), although `TODO` is mentioned for inline comments 🤷‍

andygrunwald: Should i change it or do you say "This is fine for now, next time use `@todo` in docblocks and `// TODO` in inline comments?

M4LuZ: I would suggest the second, given that this is a really minor point ;)

PR #165, opened 2018-03-21
Rework of gd class: Reformatting, PHPDoc for methods, removed dead code, removed unused methods

andygrunwald:
- Removed dead code
- Removed unused methods `Colorize` and `MergeImages`

### DIFF
--- a/inc/classes/class_gd.php
+++ b/inc/classes/class_gd.php
@@ -1,31 +1,64 @@
 <?php
+
 class gd
 {
+    /**
+     * @var resource
+     */
     public $img;
+
+    /**
+     * @var int
+     */
     public $height;
+
+    /**
+     * @var int
+     */
     public $width;
+
+    /**
+     * @var string
+     */
     public $font;
+
+    /**
+     * @var int
+     */
     public $font_size;
+
+    /**
+     * @var int
+     */
     public $free_type = 0;
+
+    /**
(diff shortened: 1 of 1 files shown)


inc/classes/class_gd.php
```diff
@@ -301,42 +382,7 @@ public function CreateThumb($old_file, $new_file, $max_width, $max_height)
...
-        $layover = imagecreate($this->width, $this->height);
-        $fill = imagefill($layover, 0, 0, imagecolorallocate($layover, $r, $g, $b));
-
-        // Merge Layover with original image
-        $merge = imagecopymerge($this->img, $layover, 0, 0, 0, 0, $this->width, $this->height, $percentage);
-        imagedestroy($layover);
-    }
-
 
-    public function MergeImages($source_file1, $source_file2, $target_file)
```
M4LuZ: I would argue that removing this is fine, as:
* imagecopymerge does not support the alpha channel correctly. Merging two pictures with transparency thus would never work as expected.
* Also merging just parts does not work, as coordinates for merge cannot be passed.

I do see the need for a similar function, but that must support the two requirements above

andygrunwald: 👍

andygrunwald:
@M4LuZ Feel free to merge it once you are ready or should i do it?

PR #167, opened 2018-03-22
Rework of func class: Reformatting, PHPDoc for methods, removed dead code, removed unused methods

andygrunwald:
- Removed dead code
- Removed unused methods

### DIFF
--- a/inc/classes/class_func.php
+++ b/inc/classes/class_func.php
@@ -1,53 +1,42 @@
 <?php
 
-/**
- * Global Functions
- *
- * @package lansuite_core
- * @author ls_admin
- * @version $Id$
- * @access public
- * @todo Remove Dialogfunctions and create own class
-*/
-
 class func
 {
-    public $ActiveModules = array();
-  /**
-   * CONSTRUCTOR : Get referer and transform in a internal Link
-   *
-   */
+    /**
+     * @var array
+     */
+    public $ActiveModules = [];
+
     public function __construct()
     {
         define('NO_LINK', -1);
+
         $url_array = [];
         $this->internal_referer = 'index.php';
 
         if (isset($_SERVER['HTTP_REFERER'])) {
             $url_array = parse_url($_SERVER['HTTP_REFERER']);
         }
+
         if (isset($url_array['query']) && $url_array['query']) {
             $this->internal_referer .= '?' . $url_array['query'];
         }
+
         if (isset($url_array['fragment']) && $url_array['fragment']) {
             $this->internal_referer .= $url_array['fragment'];
         }
     }
 
-
-  /**
-   * Read the Config-settings from DB
-   * @global mixed Databaseobject
-   * @global array Baseconfig from File
-   * @return $cfg
-   */
+    /**
+     * Read the Config-settings from DB
+     *
+     * @return array
+     */
     public function read_db_config()
     {
         global $db;
         $cfg = [];
 
-        // Idea: Select current mod only, does not work with plugin.
-        // $res = $db->qry('SELECT cfg_value, cfg_key, cfg_type FROM %prefix%config WHERE cfg_module = "install" OR cfg_module = %string%', $mod);
         $res = $db->qry('SELECT cfg_value, cfg_key, cfg_type FROM %prefix%config');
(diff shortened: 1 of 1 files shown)


inc/classes/class_func.php
```diff
@@ -200,67 +195,56 @@ public function unixstamp2date($func_timestamp, $func_art)
...
-          return $func_date;
-    }
 
-  /**
-   * Calculates the age to a given birthday timestamp
-   * @timestamp birthhday
-   * @optional timestamp attime Calculates the age the person will have at this time
-   * @return int age
-   */
-    public function age($birthday, $attime = 0)
```
M4LuZ: I'm wondering:
* Why a user-age related function is in `func` and not in `user`
* Why this is not used anywhere, although there has to be functionality to evaluate the minimum age. Duplicated code?
* Why [DateTime::diff](http://php.net/manual/de/datetime.diff.php) isn't used to provide this. Guess this is pre-PHP-5.3 developed code.

I agree with the removal from `func`, but would suggest to have a look at the current age verification system, if a similar function in user exists (or should be created)

inc/classes/class_func.php
```diff
@@ -396,7 +449,6 @@ public function text2html($string, $mode = 0)
             }
 
             if ($mode != 2) {
-                $img_start = "<img src=\"design/".$auth["design"]."/images/";
```
M4LuZ: Why was this removed?

inc/classes/class_func.php
```diff
@@ -561,21 +621,10 @@ public function Entity2Uml($string)
         return $string;
     }
 
-
-  #### Check Var Content ####
-    public function check_var($var, $type, $min_length, $max_length)
```
M4LuZ: Fine to remove. 
There is native PHP support with the `is_<type>()` functions, and the error message does not really provide added value, does it? ;)

inc/classes/class_func.php
andygrunwald: Because the variable `$img_start` is not in use at all.
In this line, it gets written, but no one (in the complete file) is accessing/reading this variable.
See https://github.com/lansuite/lansuite/blob/master/inc/classes/class_func.php#L399

inc/classes/class_func.php
```diff
@@ -594,99 +644,127 @@ public function checkIP($ip)
...
+     * @param int       $max_entries_per_page
+     * @param int       $overall_entries
+     * @param string    $working_link
+     * @param string    $var_page_name
+     * @return array
+     */
     public function page_split($current_page, $max_entries_per_page, $overall_entries, $working_link, $var_page_name)
     {
         if ($max_entries_per_page > 0 and $overall_entries >= 0 and $working_link != "" and $var_page_name != "") {
-            if ($current_page == "") {
```
M4LuZ: Reason for removal?

inc/classes/class_func.php
andygrunwald: Exactly. Once we are getting more and more stable we can introduce PHP native type system here.

inc/classes/class_func.php
andygrunwald: Here it will be checked if `$current_page` is empty. If empty, then the vars `$page_sql`, `$page_a` and `$page_b` will be set.
A couple of lines after `$current_page` will be checked again and has an `else` part.
In this `else` part the variables `$page_sql`, `$page_a` and `$page_b` will be overwritten.

See https://github.com/lansuite/lansuite/blob/master/inc/classes/class_func.php#L638-L651

As a result: This is dead code.

inc/classes/class_func.php
M4LuZ: Ah, understood!

inc/classes/class_func.php
M4LuZ: There is already an option to define a minimum age for a party and AFAIK this is also enforced on signup, so similar code must already exist at another place. But I would also need to search for it first.

But as said not blocking the removal here, may just be worth a look.

inc/classes/class_func.php
M4LuZ: Indeed, good find!

PR #170, opened 2018-03-25
Rework of display class: Reformatting, PHPDoc for methods, removed dead code, removed unused methods, fixes undefined variable bugs

andygrunwald:
- Removed dead code
- Removed unused methods
- Fixed a couple of bugs where a variable was not used or misspelled

### DIFF
--- a/inc/classes/class_display.php
+++ b/inc/classes/class_display.php
@@ -1,34 +1,70 @@
 <?php
 
+/**
+ * Class display
+ *
+ * When handling own templates, the prefered way is to use $smarty->fetch().
+ * However, at some point, you have to append these fetched content to $MainContent
+ * But: Never write to $MainContent within your module!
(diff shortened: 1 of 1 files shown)


inc/classes/class_display.php
```diff
@@ -136,142 +204,99 @@ public function AddHeaderMenu($names, $link, $active = null)
...
         }
         $items = substr($items, 0, -3);
-    
+
         $MainContent .=  $items;
     }
-  
-  // dynamic = 1 --> Jquery-Javascript-Tabs
-  // dynamic = 0 --> static tabs without Jquery & Javascript
-    public function AddJQueryTabsStart($id, $dynamic = null)
```
M4LuZ: I was surprised to see that this is considered dead code, as I saw tabs being used in multiple locations.
But it seems that this was now covered by `StartTabs()`and `StartTab()` and a bit of JS-Code in `scripts.js`

PR #176, opened 2018-03-26
Rework of auth class: Reformatting, PHPDoc for methods, removed dead code

andygrunwald:
- Removed dead code
- Removed unused methods

### DIFF
--- a/inc/classes/class_auth.php
+++ b/inc/classes/class_auth.php
@@ -1,34 +1,102 @@
 <?php
 
 /**
- * Authorisation and Cookiemanagement for Lansuite
+ * Class auth
  *
- * @package lansuite_core
- * @author bytekilla
- * @version $Id$
- * @access public
+ * Authorisation and Cookie management for LanSuite
+ *
  * @todo Change uniqkey from md5(password) to an extra Field
  */
 class auth {
 
-    public $auth = array();                 // Userdaten im Array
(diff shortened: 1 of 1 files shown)


inc/classes/class_auth.php
```diff
@@ -1,34 +1,102 @@
...
+     * @var bool
+     */
+    public $cookie_crypt = true;
+
+    /**
+     * Passphrase for AzDGCrypt
+     *
+     * @var string
+     */
+    public $cookie_crypt_pw = "iD9ww32e";
```
M4LuZ: Ahem.
This we should definitely change to an unique key per installation.
Not as a part of this, PR but in general

andygrunwald:
Agree. Was surprised to see this.
Will do in one of the next PRs

M4LuZ <notifications@github.com> schrieb am Di. 27. März 2018 um 09:11:

inc/classes/class_auth.php
```diff
@@ -575,39 +648,19 @@ public function set_cookie_pw($userid)
         $this->cookie_set();
     }
 
-  /**
-   * Validate Usercookie
-   *
-   * @return int Return the Vailidity. 1=OK, 0=NOK
-   * @access private
-   */
-    public function cookie_valid()
```
M4LuZ: This isn't in use?
Thats...disturbing

Issue #96, opened 2017-12-07
TLD .bayern domain not accepted


lbayerlein:
## Expected Behavior
Register with domain.bayern without an error

## Current Behavior
Top level domain .bayern not accepted

## Steps to Reproduce (for bugs)

1. Register on a fresh installation on lansuite with a .bayern domain

## Your Environment
It is an default behaviour and not only in my environment.

andygrunwald:
Thanks for the report. Indeed, new TLDs are not (yet) supported. This needs to be fixed asaps.

M4LuZ:
We did merge support for checks against the official TLD-list a while ago, but this is only used for email address verification so far.
See 
https://github.com/lansuite/lansuite/commit/8b4959bdf4894757cde2c899ddcad97ef3b61c1a
https://github.com/lansuite/lansuite/commit/b550e4764714957961e7ff4d9fe497cd018a5292
I guess it would also make sense to include a refresh on install to make sure the referenced list is up to date.

andygrunwald:
@M4LuZ Why we are doing this at all? Any special reason for it? What use case it is for?

M4LuZ:
Why: To fix the existing address verification, as it had an incomplete and outdated list of TLDs to check against. 
Instead of just adding whatever was needed in that case I suggested to go all the way and make it future proof by getting the most current TLD list directly from IANA.  
As said this would also work for stuff like .bayern, .nrw and whatever somebody comes up with in the future.
Usecase: Avoid (at least some) typing mistakes.

andygrunwald:
I would vote for dropping the specific list of TLDs at all and use standard verification mechanism to check the email.
Why? In the future more and more TLDs will come up. If someone want to run a version that is older that the new TLDs, but want to run this under a new TLD this might raise problems for the user.

What is the alternative?
I would vote for the need to enter the Address / E-Mail-Address twice. This lowers the risk of typos. Further more we can use libs like [symfony/validator](https://github.com/symfony/validator) ([docs](https://symfony.com/doc/current/components/validator.html)) to check the validation of the address / e-mail address here.

This behaviour is pretty standard in the internet.

With this solution we

* are able to drop the tld mysql table at all
* reduce the need of a constant cronjob to update the tlds
* make it easier for the user to run whatever TLD he/she wants

Even better: If someone is running a custom tld inside their network (i know some companies who are doing this for internal services), this would be possible with the approach suggested in this comment. With the IANA list, it would not.

[...]

PR #186, opened 2018-04-03
Fix 96 email verification

M4LuZ:
Change as agreed in #96:
* Remove TLD verification functionality
* Add field to re-type email and compare both inputs


andygrunwald:
This is a good basement! Thank you.
As I wrote in #96 I would even go one step further and make the email check a little bit more sophisticated in the future. But this fixes the issue. Thanks Marius!

lbayerlein:
Hi @andygrunwald,
I am a some busy right now. I try to check this week. Will give you feedback then. Thanks

Issue #127, opened 2018-01-16
Install on web-server with external db-server


mbraun81:
Why does the DB have to run on the web server?
The first page of the install script checks if there is a local mysql server. I think it will be better to check only if a php mod for database connection is available

andygrunwald:
The database don't have to run on the same server as the Web server.
You are able to place a configuration file into `./inc/base/config.php `.
The content is described here: https://github.com/lansuite/lansuite#configuration-file
The key `server` described the location of the server. E.g. an ip address.

If no configuration file can be found, standard configurations will be applied.
The standard configuration implies the MySQL server on localhost.
See https://github.com/lansuite/lansuite/blob/master/index.php#L218

Does this explain the behaviour? Does this help you?

mbraun81:
Maybe then the connection of the local db on the first page of the installation should not be checked. The time at which you can enter the data would be very good for.

M4LuZ:
Just raised a pull request for this.
It does still make sense to try to connect to the local DB instance
As that is the default setup and we're depending on that connection to check the MySQL server version. But it is certainly true, that it should not cause any issues if that initial connection fails.

(Although in theory it should be possible to determine the MySQL version without any credentials. Version number is contained in the initial packet sent by the server. Just telnet to port 3306 and have a look. But who wants to implement that?...)

andygrunwald:
> (Although in theory it should be possible to determine the MySQL version without any credentials. Version number is contained in the initial packet sent by the server. Just telnet to port 3306 and have a look. But who wants to implement that?...)

Nice idea, but i doubt if we really need this feature. Is the information of the MySQL version really this important to put this complexity into the source code?

PR #187, opened 2018-04-03
Fix 127 mysql localhost check

M4LuZ:
Implementation of request described in #127 :
* LanSuite still tries to connect to local MySQL, but errors for that case are surpressed
* Environment check now has a case when the server is not available
* Also added checks to compare for MariaDB-Versions, as that is now standard amongst some distributions
* Wrapped `getServerInfo()` into an if-case to check if connection is established. This was the function throwing the original problem


andygrunwald:
@M4LuZ Looks good to me.
I made a couple of small changes:

* https://github.com/lansuite/lansuite/pull/187/commits/fb4937f56688d0fab44bc16cbc6a18b7e5abd6fa: I moved the control error operator to the function because we want to suppress the error of the function, not of the assignment
* https://github.com/lansuite/lansuite/pull/187/commits/5bcbfdcf7db0b99bb1b42c1cdeb0faf73fb6a0eb: I used the `link_id` variable because this is `false` once the connect don't work. With this, we don't need to maintain a second variable (`success`). Later we can drop the success variable because the `link_id` reflects the same in our use case

I approved this PR. When you agree with my changes, feel free to hit the merge button yourself.

PR #188, opened 2018-04-03
fix #142 Tournament overview not displayed if played as league

### LINKED ISSUE #142: Tournament overview not displayed if played as league
Originally reported on LS page: http://lansuite.orgapage.de/index.php?mod=board&action=thread&tid=1392&posts_page=0#pid7770

If a tournament is created and the mode is league then the detailed overview of the matches contains no data.  Neither team names nor match results are shown.
## Expected Behavior
Display of team names on vertical and horizontal top/left row/column
Display of match results in the table.
Link to match details per table entry
## Current Behavior

Real-Life example here: 
https://berg-lan.de/index.php?mod=tournament2&action=tree&step=2&tournamentid=45
If a tournament is created and the mode is league then the detailed overview of the matches contains no data.  Neither team names nor match results are shown.

## Possible Solution
To-be investigated. Simplest solution: Replace by simple table

1. Create tournament with mode = league
2. Play a few matches
3. Look at the match overview

## Context

Implications are only a loss of comfort, as the match listing still works.

## Your Environment

* Version used: maluz/lansuite/HEAD (Basically 4.2 with a few fixes)
* Operating System and version: Debian 8
* Enabled features: tournament2

M4LuZ:
Issue and workaround described in #142 
Issue seems to be related to the group selection criteria.
tfunc->GetTeamAnz determines the right value per default, but as an empty string is passed along (from tree.php to tree_frame.php to tfunc->GetTeamAnz()), it tries to use that instead, ending up in the empty display


andygrunwald:
@M4LuZ I made a change to this PR in https://github.com/lansuite/lansuite/pull/188/commits/91d05bf639c6d54a3215b38563f650e82c4e69e9

The `$_POST['group']` parameter contains a `"0"` (string) once the "Finalspiele" was chosen in the dropdown build here: https://github.com/lansuite/lansuite/blob/master/modules/tournament2/tree.php#L60-L63

Due to the `(int)` cast + string conversion it ends up in an empty string, which overwrite the default value `0` or the `GetTeamAnz` method.
But once a different value was chosen in the select, it should pass correctly the group id into the `tree_frame`.

My change handles both cases. I approved the PR. If you agree here, feel free to merge it.

PR #190, opened 2018-04-11
Limit API of inc/classes while switching visibility for properties and methods

andygrunwald:
LanSuite was written when PHP didn't support any kind of visibility.
When we started to work on LanSuite again, we made everything public (which was the default). This commit switches the visibility from public to private for properties and methods that are not accessed
by other entities.

This change allows us to define a more granular API, refactoring and testing.

### DIFF
--- a/inc/classes/class_auth.php
+++ b/inc/classes/class_auth.php
@@ -14,70 +14,70 @@ class auth {
      *
      * @var array
      */
-    public $auth = [];
+    private $auth = [];
 
     /**
      * Time
      *
      * @var int
      */
-    public $timestamp;
+    private $timestamp;
 
     /**
      * Cookie data
      *
      * @var array
      */
-    public $cookie_data = [];
+    private $cookie_data = [];
 
     /**
      * Cookie name
      *
      * @var string
      */
-    public $cookie_name = 'LSAUTH';
+    private $cookie_name = 'LSAUTH';
 
     /**
      * Cookie version
      *
      * @var string
      */
-    public $cookie_version = '1';
+    private $cookie_version = '1';
 
     /**
      * Domain
      *
      * @var string
      */
-    public $cookie_domain = '';
+    private $cookie_domain = '';
 
     /**
      * Duration in days
      *
      * @var string
      */
-    public $cookie_time = '30';
+    private $cookie_time = '30';
 
     /**
      * Cookie path
      *
      * @var string
      */
-    public $cookie_path = '';
+    private $cookie_path = '';
 
     /**
      * Crypt Cookie with AzDGCrypt
      *
      * @var bool
      */
-    public $cookie_crypt = true;
+    private $cookie_crypt = true;
 
     /**
      * Passphrase for AzDGCrypt
      *
      * @var string
      */
-    public $cookie_crypt_pw = "iD9ww32e";
+    private $cookie_crypt_pw = "iD9ww32e";
 
     /**
      * Array containing all users, currently online
@@ -379,7 +379,7 @@ class auth {
      * @param string $uniquekey
      * @return void
      */
-    public function login_cookie($userid, $uniquekey)
+    private function login_cookie($userid, $uniquekey)
     {
         global $func;
 
@@ -585,7 +585,7 @@ class auth {
      *
      * @return mixed
      */
-    public function loadAuthBySID()
+    private function loadAuthBySID()
     {
         global $db;
         // Put all User-Data into $auth-Array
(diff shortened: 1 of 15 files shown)


inc/classes/class_auth.php
```diff
@@ -14,70 +14,70 @@ class auth {
      *
      * @var array
      */
-    public $auth = [];
+    private $auth = [];
```
M4LuZ: There is a reference to this in `index_module.inc.php` (L32)

inc/classes/class_auth.php
```diff
@@ -635,7 +635,7 @@ public function update_visits($frmwrkmode = "")
      * @param int $userid
      * @return void
      */
-    public function set_cookie_pw($userid)
+    private function set_cookie_pw($userid)
```
M4LuZ: Is called in usrmgr/changepw.php line 22. Would also cause a problem.

inc/classes/class_auth.php
andygrunwald: Oh, true. Fixed.

inc/classes/class_db_mysql.php
```diff
@@ -5,17 +5,17 @@ class db
     /**
      * @var int
      */
-    public $link_id = 0;
+    private $link_id = 0;
```
M4LuZ: used in class_import.php line 808

inc/classes/class_auth.php
andygrunwald: My IDE didnt catched this because of all the global variables. Fixed.

inc/classes/class_db_mysql.php
andygrunwald: This is also changed in this PR. See https://github.com/lansuite/lansuite/pull/190/files#diff-07003def8a09985dfaa2f74d4cbcb1a2

inc/classes/class_db_mysql.php
```diff
@@ -75,7 +75,7 @@ public function print_error($msg, $query_string_with_error)
      * @param array $match
      * @return int|mixed|string
      */
-    public function escape($match)
+    private function escape($match)
```
M4LuZ: I can't find any call to this function. Would vote for removal instead

inc/classes/class_debug.php
```diff
@@ -207,7 +207,7 @@ function compare($wert_a, $wert_b) {
      * @param string $name  Table heading
      * @return string       HTML-Row for table (<tr><td>...</td></tr>)
      */
-    public function row_top($name)
+    private function row_top($name)
```
M4LuZ: This function also seems to be unused

inc/classes/class_db_mysql.php
andygrunwald: This is in use here: https://github.com/lansuite/lansuite/blob/master/inc/classes/class_db_mysql.php#L193

inc/classes/class_masterform.php
```diff
@@ -40,42 +40,32 @@ class masterform
...
     /**
      * @var array
      */
-    public $WYSIWYGFields = [];
+    private $WYSIWYGFields = [];
 
     /**
      * @var array
      */
-    public $SQLFieldTypes = [];
```
M4LuZ: SQLFieldTypes is used within class_mastersearch2.php and thus declaration shouldn't be removed

inc/classes/class_translation.php
```diff
@@ -393,7 +393,7 @@ public function xml_read_to_array($module) {
      * @param string    $module     Module name (System, DB, Module ...)
      * @return string
      */
-    public function get_trans_filename($module)
+    private function get_trans_filename($module)
```
M4LuZ: Referened in TranslationTest.php L160

inc/classes/class_xml.php
```diff
@@ -149,7 +149,7 @@ public function get_tag_content($tag_to_grab, $input, $replace_html = true) {
      * @param int       $level
      * @return string
      */
-    public function level2tab($level) {
+    private function level2tab($level) {
```
M4LuZ: Referenced in XMLtest.php L46

inc/classes/class_translation.php
andygrunwald: Deleted in this PR also: https://github.com/lansuite/lansuite/pull/190/files#diff-ce67f17604206546cec768ee5f8b7802L158

inc/classes/class_xml.php
andygrunwald: Deleted in this PR also: https://github.com/lansuite/lansuite/pull/190/files#diff-5509b988ae8d72113b9cace28e758387L44

inc/classes/class_db_mysql.php
M4LuZ: well....shit. Did only a search for function calls, Not sure how we can search for callback uses :/

PR #198, opened 2018-04-11
class auth: Reset login key in auth storage

andygrunwald:
The expression `==` is doing a comparison. As a result, a bool is returned. But this value is never caught.
This seems to be like a bug. The original functionality seems to be a reset of the auth array


M4LuZ:
I would vote to remove the entry instead. 
Fixing the assignment would modify functionality and would require to check if every read operation still yields the expected result, while removing would ensure the same functionality.
Unless you did already check that?

andygrunwald:
Yes, i checked most of the calls.
The are like
```
if ($auth['login']) {
...
}
```

So only checks on the boolean value here.
In this case it would not change the functionality. But i saw that fixing this don't make sense, because the `login` field is resetted a couple of lines above.
See https://github.com/lansuite/lansuite/blob/master/inc/classes/class_auth.php#L408-L418

PR #200, opened 2018-04-11
class display: Remove function AddIcon and replace it with FetchIcon

andygrunwald:
`AddIcon` was marked as deprecated a couple of month ago.
The function itself calls `FetchIcon`. This PR replaces `AddIcon` with `FetchIcon`


modules/hardware/plugins/usrmgr_details_tab.php
```diff
@@ -8,25 +8,25 @@
...
-$dsp->AddDoubleRow(t('Festplatte 1'), $dsp->AddIcon('hdd', '', '').' '.$hardware['hdd1']);
-$dsp->AddDoubleRow(t('Festplatte 2'), $dsp->AddIcon('hdd', '', '').' '.$hardware['hdd2']);
-$dsp->AddDoubleRow(t('Optisches Laufwerk 1'), $dsp->AddIcon('cd', '', '').' '.$hardware['cd1']);
-$dsp->AddDoubleRow(t('Optisches Laufwerk 2'), $dsp->AddIcon('cd', '', '').' '.$hardware['cd2']);
-$dsp->AddDoubleRow(t('Maus'), $dsp->AddIcon('maus', '', '').' '.$hardware['maus']);
-$dsp->AddDoubleRow(t('Tastatur'), $dsp->AddIcon('tasta', '', '').' '.$hardware['tasta']);
-$dsp->AddDoubleRow(t('Monitor'), $dsp->AddIcon('screen', '', '').' '.$hardware['monitor']);
-$dsp->AddDoubleRow(t('Betriebssystem'), $dsp->AddIcon('os', '', '').' '.$hardware['os']);
-$dsp->AddDoubleRow(t('Computername'), $dsp->AddIcon('pc', '', '').' '.$hardware['name']);
+$dsp->AddDoubleRow(t('CPU'), $dsp->FetchIcon('', 'cpu', '').' '.$hardware['cpu']);
```
M4LuZ: General remark: Wouldn't it have been better to switch the first arguments in the functional definition of FetchIcon to both avoid to modify every function call and to leverage the assignment of default values to trim the function calls down the one variable?

andygrunwald: Valid point. But I count ~83 usages of `FetchIcon`  method. So we had to change those, if we would switch the order of the arguments.
I think it doesn't matter in this case.

M4LuZ: > I think it doesn't matter in this case.

Quite surprised to hear that from from the master of smaller improvements 😛 
I would say this offers an usability improvement and should be done.
Shouldn't be too hard to do with some RegEx magic, I might even give it a shot myself...

andygrunwald: Feel free :)

andygrunwald: I saw that my IDE had a refactor feature for this usecase.
So i did it in a couple of minutes.

andygrunwald:
@M4LuZ Ready for review again.

modules/hardware/plugins/usrmgr_details_tab.php
M4LuZ: Cool. What IDE is that?

andygrunwald: https://www.jetbrains.com/phpstorm/

PR #202, opened 2018-04-11
Fix #96: Add more detailed email validation check

### LINKED ISSUE #96: TLD .bayern domain not accepted

andygrunwald:
This PR is a follow up from #186 for #96.
This adds a little bit more granular checks for emails incl. basic DNS checks.

Furthermore, the email functionality is backed by unit tests.

This PR adds three new configuration options:

```ini
[validation]
email[mode] = "html5"
email[mx_check] = 1
email[host_check] = 0
```

* mode: "html5" (W3C regular expression) or "loose"
* mx_check: MX DNS Record will be checked
* host_check: MX, A or AAAA DNS record will be checked

Related ticket: #96


inc/classes/Validator/Email.php
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * Check if one of MX, A or AAAA DNS RR exists.
+     *
+     * @param string $host
+     * @return bool
+     */
+    private function checkHost(string $host): bool
+    {
+        return '' !== $host && ($this->checkMX($host) || (checkdnsrr($host, 'A') || checkdnsrr($host, 'AAAA')));
```
M4LuZ: I suspect that this would completely disable registration for any system that cannot reach a DNS resolver (misconfiguration, connectivity issues or local installation without internet access) as the return value FALSE  of `checkdnsrr` is set for both technical issues and missing records.
Would it make sense to use [`dns-get-record()`](http://php.net/manual/de/function.dns-get-record.php) instead?

M4LuZ: Ahhh, you got me there!
Misread the documentation, I thought it would also return something as array if the lookup fails.
So no benefit from using a different built-in function.
There are pure PHP implementations (e.g. [NetDNS2](https://github.com/mikepultz/netdns2)) that throw useful exceptions, but I suspect that would be over-engineering a simple problem.

I would argue against the two configuration settings though.
Wouldn't it make more sense to put these as setting into of the usermgmt-module or general system config? (maybe even as general "enable/disable DNS lookups"?)

We could also simply define that the DNS verification only blocks registration on internet-systems and only shows a warning for intranet-systems. 
DNS configuration should then be checked during the installation.

What do you think?

andygrunwald: > but I suspect that would be over-engineering a simple problem
I agree here.

> Wouldn't it make more sense to put these as setting into of the usermgmt-module or general system config? (maybe even as general "enable/disable DNS lookups"?)
I agree. From the first glance, this would make sense. Let me have a look later this day how we can implement this.

> We could also simply define that the DNS verification only blocks registration on internet-systems and only shows a warning for intranet-systems.
True in general. But i would split this feature and not include it here in this PR.
Why? Because i think we have multiple places that don't work in an intranet env (e.g. mails).
How is your view here?

M4LuZ: >True in general. But i would split this feature and not include it here in this PR.
Why? Because i think we have multiple places that don't work in an intranet env (e.g. mails).
How is your view here?

Simple: There is already an option to define if an installation is for internet/intranet (`$cfg["sys_internet"]`) which is used to switch functionality in a lot of places. Thus that does luckily not require any additional work

UPGRADE-5.0.md
```diff
@@ -0,0 +1,24 @@
...
+Many LanSuite installations have database table prefixes.
+We assume that your database table prefix is `ls_`.
+If this is different for your installation, please change the table names in the queries accordingly.
+
+ * The column `cfg_key` in table `config_selections` was extended from `CHAR(16)` to `VARCHAR(50)`
+
+   To upgrade, execute the query
+
+   ```sql
+   ALTER TABLE `ls_config_selections` MODIFY `cfg_key` VARCHAR(50);
```
M4LuZ: Changes to the DB structure based on the XML definition can be automatically applied with the DB manager.

andygrunwald: True.
I thought this might be useful for people who modified their instance. But i might think to much.
Do you still think this file make sense? If yes, we can describe this here, that the changes can be done via the Admin installation module.
Or maybe useful for people who wrote their own modules.
Or is this too much? Whats your suggestion? Removing or keeping this file? And when keeping, which purpose should this fulfill?

M4LuZ: A general update documentation(per release) is a good idea, also started one in the wiki a while ago:
https://github.com/lansuite/lansuite/wiki/HowTo%3A-Upgrade-from-Release-4.2
I would only document the queries that cannot be covered by the existing db update (e.g. necessary DML statements or preparative / rollback steps), because the rest can be seen either by a diff of the XML documents or by just looking at the generated DDL statements on execution

andygrunwald: Fine for me. Removed the file again and created https://github.com/lansuite/lansuite/issues/214.
Feel free to continue with the review.

PR #217, opened 2018-04-15
Move of inc/classes into LanSuite namespace

andygrunwald:
What this  PR does:

- Renamed inc/classes to inc/Classes to match a pattern with the Functions folder and the Classes folder in the modules
- Renames the files in inc/Classes to match the PSR-PHP-Standard pattern
- Moved all classes in inc/Classes to LanSuite namespace
- Renamed class `auth` to `Auth`
- Renamed class `barcode` to `Barcode`
- Renamed class `barcode_system` to `BarcodeSystem`
- Renamed class `db` to `DB`
- Renamed class `debug` to `Debug`
- Renamed class `display` to `Display`
- Renamed class `framework` to `Framework`
- Renamed class `func` to `Func`
- Renamed class `gd` to `GD`
- Renamed class `Mastercomment` to `MasterComment`
- Renamed class `masterdelete` to `MasterDelete`
- Renamed class `masterform` to `MasterForm`
- Renamed class `masterrate` to `MasterRate`
- Renamed class `plugin` to `Plugin`
- Renamed class `sec` to `Security`
- Renamed class `translation` to `Translation`
- Renamed class `xml` to `XML`


andygrunwald:
Yes. I did a completely new installation.
After this i clicked around, created a news and added a new party.

This change is so big, if this didn't worked out, the only thing you can see is a Fatal Error, because we would not be able to load _any_ class.
So this is quite obvious if this works or not.

> Also: Please add the class and folder renaming to the update documentation, as this will break compatibility with any module not maintained in this repo

Good point! One suggestion: Once this is merged, I will go through all the commits since the 4.2 release and update it once to the point where we are right now. In a separate PR. What do you think here?

andygrunwald:
> That would be the case if autoloading fails, but not if one of the includes or class instantiations was missed in the source code. 

If a `new X` was missed, a Fatal error is thrown as well.
If a `class::constant` is called, the same. Fatal error is thrown, because the class is not known.

I would go with a _move fast, fail fast_ approach.
Once a Fatal error is known, it is quite easy to fix: Refer to the new class incl. the namespace.
It is a small change and can be applied after.
Furthermore, such cases can be found due to static analyzing, because all classes are autoloaded and then the missing reference would lead to a "unknown class".

andygrunwald:
@M4LuZ Any update here? Anything I can improve to make it easier for you to review this?

andygrunwald:
Conflict resolved. When the tests are green, we can continue here.

andygrunwald:
👍

PR #218, opened 2018-04-28
Module cleanup: Removed silly assignments, unreachable code and more

andygrunwald:
This PR doesn't contain functionality changes.
It is only about code cleanup:

* Removed silly assignments (where both sides are the same)
* void return values used
* Removed unnecessary javascript label
* Removed unreachable statement

### DIFF
--- a/modules/server/ping_server.inc.php
+++ b/modules/server/ping_server.inc.php
@@ -20,7 +20,7 @@ function ping_server($host, $port)
         $available = $success;
 
         $special_info = "";
-        // Weitere Daten f�r FTPs herrausfinden
+        // Weitere Daten für FTPs herrausfinden
         if (($success) && ($server_daten["type"] == "ftp")) {
             if ($fp = @fsockopen($host, $port, $errno, $errstr, 1)) {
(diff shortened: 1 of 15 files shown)


modules/server/ping_server.inc.php
```diff
@@ -20,7 +20,7 @@ function ping_server($host, $port)
         $available = $success;
 
         $special_info = "";
-        // Weitere Daten f�r FTPs herrausfinden
+        // Weitere Daten f�r FTPs herrausfinden
```
M4LuZ: Encoding strikes again ;)

modules/server/ping_server.inc.php
```diff
@@ -85,7 +85,7 @@ function ping_server($host, $port)
         } // END: If Type=FTP
 
 
-        // Weitere Daten f�r IRCs herrausfinden
+        // Weitere Daten f�r IRCs herrausfinden
```
M4LuZ: dito

andygrunwald:
@M4LuZ Gimme a second.

andygrunwald:
@M4LuZ Fixed.

PR #219, opened 2018-04-28
Module clean: Removed unused local variables

andygrunwald:
This PR doesn't contain functionality changes.
It is only about code cleanup:

* Removed unused local variables


modules/install/class_import.php
```diff
@@ -47,8 +47,7 @@ public function GetImportHeader($tmp_file_name)
     // See: http://lansuite.orgapage.de/index.php?mod=bugtracker&bugid=1059
     #$this->xml_content = str_replace("''", "'", $this->xml_content);
 
-        ## Get Header-Tag
-        $header = "";
+        // Get Header-Tag
         $this->xml_content_lansuite = $xml->getFirstTagContent("LANsurfer", $this->xml_content, 0);
```
M4LuZ: I think we can remove the whole LANsurfer code. Project died in 2006 as far as I can see. Not sure anybody would need an import function still. And if they do: they can import with LanSuite 4.2 and then move forward from there

andygrunwald: Will do. Let me create a ticket for this and handle this in a separate PR.

modules/signon/class_signon.php
```diff
@@ -80,10 +80,9 @@ public function GetLansurfer($username, $password)
             fclose($handle);
 
             // Log out
-            $handle = fopen("http://lansurfer.com/user.phtml?action=logout?LS_Session=$LS_Session", 'r', false, stream_context_create($opts));
+            fopen("http://lansurfer.com/user.phtml?action=logout?LS_Session=$LS_Session", 'r', false, stream_context_create($opts));
```
M4LuZ: See above. Would remove code related to lansufer completely.

modules/install/class_import.php
andygrunwald: @M4LuZ As a follow up: https://github.com/lansuite/lansuite/issues/222

modules/signon/class_signon.php
andygrunwald: See https://github.com/lansuite/lansuite/issues/222

andygrunwald:
@M4LuZ Conflict resolved.

andygrunwald:
Conflict resolved. When the tests are green, we are good to go.

PR #223, opened 2018-04-29
Move global functions to composer autoloader

andygrunwald:
It moves the global functions

- CheckValidEmail()
- d()
- dependencyPreCheck()
- FetchDataRow()
- FetchPostRow()
- MasterCommentEditAllowed()
- t() and t_no_html()

to the composer autoloader


andygrunwald:
Conflict resolved. When the tests are green, this should be good to go.

Issue #214, opened 2018-04-14
Update documentation per Release


andygrunwald:
## Idea

Quote by @M4LuZ from https://github.com/lansuite/lansuite/pull/202#discussion_r181548838

> A general update documentation(per release) is a good idea, also started one in the wiki a while ago:
https://github.com/lansuite/lansuite/wiki/HowTo%3A-Upgrade-from-Release-4.2
I would only document the queries that cannot be covered by the existing db update (e.g. necessary DML statements or preparative / rollback steps), because the rest can be seen either by a diff of the XML documents or by just looking at the generated DDL statements on execution

## Possible Solution

Create a markdown file like the symfony/symfony project.
Example: [symfony/symfony: UPGRADE-5.0.md](https://github.com/symfony/symfony/blob/master/UPGRADE-5.0.md)

Why markdown and not the Wiki?
With this a code review can take place to ensure a) quality and b) understanding of multiple persons instead of one. Because often the one who wrote, is way to deep into the problem to describe the upgrade to people who are not aware of the details of the issue.

andygrunwald:
Maybe related: https://keepachangelog.com/en/1.0.0/

andygrunwald:
Important note: Add a paragraph about the fonts. See https://github.com/lansuite/lansuite/pull/234#issuecomment-386531735

Issue #90, opened 2017-07-18
Implement better upgrade handling


M4LuZ:
In order to allow us to facilitate various changes of lansuite, its modules and the database structure in the background we need a good way to run upgrades. The current system has some shortfalls that does not support the large changes that we need to do.

## Expected Behavior

Expected Behaviour is specified on a [dedicated wiki page](https://github.com/lansuite/lansuite/wiki/Module-Handling---Requirements)

## Current Behavior
* Module Mgr. only allows DDL changes, through changes in the module XML-files. There is no way to run self-defined queries
* There is no way to execute PHP code
* There is no way to execute additonal tasks before or after the upgrade
* There is no way to do a rollback to the previous version
* There is no way to model multiple upgrade steps or dependencies  between modules

## Possible Solution

Solution to be modelled after requirements are clear

## Steps to Reproduce (for bugs)

Try to do an upgrade ;)

## Context

We do see the first changes that require a defined update sequence to be applied successfully.
As the update process does not support that, we have to rely on the user to run tasks in the right order and handle everything that the current update process cannot handle

## Your Environment
* Version used: all of them
* Operating System and version: all of them
* Enabled features: N/A

lxp:
I like the requirements you described on the wiki page. The most important thing is to track DB versions, because otherwise we can't decide what to update at all.

As far as I know, LanSuite currently doesn't allow complex DB updates at all. It only works if the database server already does the right thing for us.

I think we might just implement rollback via DB transactions, as implementing it on our own might be too much effort and also does not bring that much additional benefit. However, this also implies that we give up some of the update process flexibility.

M4LuZ:
> The most important thing is to track DB versions, because otherwise we can't decide what to update at all.

We also need to track the source code versions (separately), because files may change without a need to upgrade the DB or the other way around. 
Or we say we use one version number to track both and the update definition says what actually needs to be done.

> I think we might just implement rollback via DB transactions,

Sadly that only works in MySQL for DML statements (INSERTS, UPDATE, DELETE), but no DDL (CREATE, ALTER, DROP), which are more likely for updates. Running any DDL statement in a transaction forces an implicit commit (so no chance of rolling back that transaction).
Given the normal size of the lansuite database, we should be fine with a pre-update DB-backup that could be restored if the upgrade fails.
[...]

PR #234, opened 2018-05-01
Fix #232: Upgrade fpdf to v1.8.1

### LINKED ISSUE #232: Module "pdf": Upgrade fpdf and include it as composer dependency
## Context

In the PDF Module we use external code  (fpdf in version 1.52).
fpdf itself is available as composer dependency in version v1.8.1
See https://packagist.org/packages/setasign/fpdf

## Expected Behavior

Upgrade fpdf to v1.8.1 and include this as composer dependency.

## Your Environment

* Version used: Current master (rev c1b2c82)
* Enabled features: pdf-module

andygrunwald:
This is a follow up of #233.

**Attention**: Don't merge this before #233. This branch depends on #233.

This PR:
- Updates fpdf to v1.8.1
- Adds fpdf as composer dependency

This PR has one disadvantage:
In the current LanSuite version, all fonts are stored in `ext_inc/pdf_fonts/`. The `ext_inc/` folder is used to store external / user data. During an update, this folder should not be touched.
In the folder `ext_inc/pdf_fonts/` you can add your custom fonts via fpdf (see [Adding new fonts and encodings](http://www.fpdf.org/en/tutorial/tuto7.htm).
The format of fonts has changed. This means that the fonts stored in `ext_inc/pdf_fonts/` can't be used anymore.
Users who added custom fonts are not able to use those anymore, because we dropped the custom font path.

My recommendation:
We document this in the upgrade path document. We write that users should drop the content of the `ext_inc/pdf_fonts/` folder and recreate their custom fonts and add them to `ext_inc/pdf_fonts/`.
With this only custom, fonts are stored in `ext_inc/pdf_fonts/`.
Native fonts are stored in vendor-folder.

@M4LuZ If you agree with this, I would continue and implement this feature for custom fonts in this PR.

Related issue: #232


M4LuZ:
> The format of fonts has changed. This means that the fonts stored in ext_inc/pdf_fonts/ can't be used anymore.
Users who added custom fonts are not able to use those anymore, because we dropped the custom font path.

Sorry, but is this now incompatible just because of the path or because the fonts themselves became incompatible?

andygrunwald:
> [...] because the fonts themselves became incompatible

This.
See e.g. the font *helvetica.php*:

- [Current helvetica.php in LanSuite](https://github.com/lansuite/lansuite/blob/master/ext_inc/pdf_fonts/helvetica.php)
- [New helvetica.php in fpdf](https://github.com/Setasign/FPDF/blob/master/font/helvetica.php)

As you see in the new version tons of more information are stored per font. According to my tests, that information is necessary to render the new font.

The thing is: The `ext_inc` folder is used to store user information. And in my opinion we should treat it like this, because this makes updates much easier for us and the user.
Because of this we 

a) need to change the font path or
b) need to explain this to every user and edit files manually there

Is this now more clear?

M4LuZ:
>The thing is: The ext_inc folder is used to store user information.

Not just that, but anything that is no source-code. Not necessarily user-related.
So I would just leave the `define('FPDF_FONTPATH', 'ext_inc/pdf_fonts/');`in.
Regarding incompatible font files:
A warning in the upgrade guide would be the minimal requirement.
If we get an upgrade function up and running (see mockup in #90), we could just check the folder for content and warn if any fonts are found there.

andygrunwald:
> Not just that, but anything that is no source-code. Not necessarily user-related.

The font-definition are PHP files and those are source code.
So they should not stay in `ext_inc`.

Related to the update, i added this to https://github.com/lansuite/lansuite/issues/214

PR #256, opened 2018-05-13
Module "hardware": Code cleanup and CI checks

andygrunwald:


modules/hardware/plugins/usrmgr_details_tab.php
```diff
@@ -1,12 +1,5 @@
 <?php
-// This File is a Part of the LS-Pluginsystem. It will be included in
```
M4LuZ: I would keep this comment in as it does explain what this does

PR #289, opened 2018-05-20
Module "teamspeak2": Rework of the module

andygrunwald:
What this PR does:

- Moved class TeamspeakDisplayClass to Classes-Folder
- Deleted CHANGELOG, because module changelogs should be reflected in the main changelog of LanSuite
- Deleted LICENSE, because LanSuite license applies on all modules
- Didn't applied namespace, because the TeamspeakDisplayClass is 3rd party code and should be extracted later on
- Applied CGL checks to TravisCI


andygrunwald:
> Fine with this PR, but I would suggest that also here we get the replacement in form of the TS3 and Discord modules ready and then remove this altogether

Agree. I think thats what #291 is for :)

M4LuZ:
No, actually it is #292 ;)
But Plan is to merge all voice server modules with #291

PR #292, opened 2018-05-22
Import new module "Teamspeak3"

M4LuZ:
Had this one lying around for quite a bit.
Provides basic functionality to integrate a TeamSpeak3 server into LanSuite.
Does support a server overview (channels and users) and allows a direct joining by creating a link.
Envisioned (and started functionality) would also allow the automated creation of tournament channels with one-time passwords.

So, to merge this the following still has to be done:
* complete migration from static inclusions to composer autoloading
* rewrite paths for images
* refactoring
* test that it still works


andygrunwald:
I had a small look at it.

- [X] Complete migration from static inclusions to composer autoloading: Done
- [X] Rewrite paths for images: Done
- [X] Refactoring: Done
- [ ] Test that it still works

There is a `Teamspeak3` class. This class is not in use at all.
Furthermore, it seems that there is code missing or is the functionality like adding channels for tournaments not implemented?

composer.json
```diff
@@ -18,7 +18,8 @@
 
         "smarty/smarty": "^3.1",
         "geshi/geshi": "^1.0.9",
-        "setasign/fpdf": "^1.8"
+        "setasign/fpdf": "^1.8",
+        "planetteamspeak/ts3-php-framework": "^1.1"
```
M4LuZ: I added this as an optional suggestion and not as requirement, as it is only required by this single module.
Similar to how the SNMP-module was handled for the NOC

M4LuZ:
As the general caching stuff is merged I would also pull this in now. hope anyone can give this a quick review?

andygrunwald:
@M4LuZ Is this still valid? Do you plan to rebase / re-pick this up and update it?
Just raising the question before I jump into a review.

andygrunwald:
@M4LuZ Is this still valid? Do you plan to rebase / re-pick this up and update it?
Just raising the question before I jump into a review.

andygrunwald:
@M4LuZ I am planning to go through the open Pull Requests and see which one we can finish/merge and get value from or which one we should abandon and close without merging.

How do you feel about this one?

I am not a user of Teamspeak, that's why I can't say much about it. Would be great to have a decision to clean up the existing PRs.

M4LuZ:
Hi @andygrunwald,

yes, I would clean-up and merge this on short notice.
the unused class (& functions) were intended to provide a more abstract and extended functionality, but I intend to cover that with a generic module (as per #828) and leave this with the current functionality that it provides

andygrunwald:
@M4LuZ Ok. Let me know once this is ready again for review.

M4LuZ:
I'm banging my head against the wall now after spending multiple hours trying to identify why inclusion of the framework through the namespace fails.

I guess that it is a stupid mistake, but I cannot find it
I did an example implementation in `show.php` based on https://github.com/planetteamspeak/ts3phpframework/tree/dev?tab=readme-ov-file#php-code-in-mvc-use-solution , but it fails on autoloading of `TeamSpeak3` in line 22.
Manual inclusion causes the framework itself to fail once it tries to instantiate the right object as it then cannot find it's own class definition for `ts3server`.

@andygrunwald: It would be great if you could have a look, as I fail to identify the issue

andygrunwald:
@M4LuZ I had a quick look at it and did:

* Included the last stable version and not the development version: https://github.com/lansuite/lansuite/pull/292/commits/d8c4ea6ae2fa3a5f6d92852dcbacb47543bd3143
* Fix inclusion of the classes: https://github.com/lansuite/lansuite/pull/292/commits/f5024c52b372e2a865d1b59042b3f3600062fbfd
* Fix syntax error in the tests: https://github.com/lansuite/lansuite/pull/292/commits/6db412ce31477744e61cd3257bc7ae611cc0adb5

It seems that `ts3server` vs. `serverquery` was the issue. I don't see `ts3server` anywhere in the documentation. It seems that `serverquery` as a scheme is the right one.

Now, the page fails with logic/connection, see

Screenshot 2024-04-08 at 13 42 55

At least, the classes are available. Does this help you to continue on this?

M4LuZ:
Yes, that does help.
I did miss the obvious, many thanks for that!

We may need to stay on `dev-master` till their next release as there is a function breaking memory leak issue with PHP 8.1 in the composer version that I did hit personally. (see https://github.com/planetteamspeak/ts3phpframework/issues/203)

M4LuZ:
Right, progress has been made while having not been made at the same time:
* return to last stable directly runs into out-of-memory issue on any connection error, thus not viable for the moment. 
The wild thing is, that `dev-master` is apparently the package before renaming to `dev-dev` and is on state of release `1.1.35`, thus originally also before PHP 8.x compatibility. 
But it works whereas the latest release on state of `dev-dev` / `1.2.1` does not?
* Namespacing issue with `dev-master` identified: There isn't any. 
Release uses PSR-4 autoloading and Namespace `PlanetTeamSpeak\TeamSpeak3Framework\` , whereas dev uses direct flle inclusion thus all framework runs in the Lansuite namespace.

I guess I have to raise a ticket with them to sort this out and pick the best tested-to-be-working version for the meantime

M4LuZ:
Just a quick FYI that this is on hold till memory leak has been solved in the framework.
Ticket raised here: https://github.com/planetteamspeak/ts3phpframework/issues/216

PR #294, opened 2018-05-23
Reset attributes of class Party back to public as workaround

M4LuZ:
Better would be to write accessor functions, replace all direct accesses and reset the class attributes back to private

See issue reported in #293


andygrunwald:
My fault. Sorry for this. And thanks for the quick fix @M4LuZ

PR #307, opened 2018-05-26
Delete test.php file

andygrunwald:
Deletes a file called `test.php` without any content.

In my opinion this should not part of the LanSuite base.
If someone wants to test PHP code, the developer can place a file themselves there.

Maybe we can put this into a documentation, but i am not sure if this it worth it.


andygrunwald:
@M4LuZ Any feedback here?

PR #308, opened 2018-05-26
Module "usrmgr": Rework of the module

andygrunwald:
What this PR does:

- Applied CGL
- Added CGL check to TravisCI
- Extracted classes to Classes/-Folder and applied namespace
- Extracted functions to Functions/-Folder
- Added PHPDocs


modules/usrmgr/add.php
```diff
@@ -394,9 +46,14 @@ function ShowFieldUsrMgr($key)
   
                     // Module-Permissions
                     $selections = array();
```
M4LuZ: can be replaced by static initialisation

PR #309, opened 2018-05-26
Module "noc": Delete module from LanSuite Codebase

andygrunwald:
This PR deleted the _Network Operation Center_ (`noc`)-Module from the LanSuite codebase.
The purpose of this module:

> Management und Statistiken der Switches (In der Testphase)

Nowadays there a lot of different network devices out there.
Often their Firmware provides statistics and management opportunities themselves.
It would be nearly impossible to offer this for all various devices.
And i think this is not the purpose of this CMS.


M4LuZ:
Here I completely and utterly disagree!
There is a ton of reasons a network operations module is very helpful.
We experienced quite a few use-cases ourselves where a good module would have helped us tremendously.
E.g.:
* Initial automated switch setup
* Configuration storage
* Mapping clients to switch ports (for troubleshooting purposes)
* Traffic and network monitoring
* On-the-fly-reconfiguration (e.g. some parties move clients to a different VLAN for tournaments)

The different variants can be covererd by good abstraction, as just a few SNMP OIDs will really be required (and are hopefully are contained in one of the standard MIBs).

andygrunwald:
That is completely fine that you have a different opinion. That is what a team is for. Discuss topics, argue and come to a better end result.

I agree with your use cases. Those sound good and useful.
But on the other side, *I* would never run this via a system like lansuite. To cover those use cases I would use tools that are specialized at this. E.g. automatic setup via [Ansible](https://www.ansible.com/). Or for traffic and network monitoring, I would use professional monitoring systems like [Prometheus](https://prometheus.io/) or [Grafana](https://grafana.com/). Prometheus has a special [SNMP Exporter](https://github.com/prometheus/snmp_exporter) for this. And those tools are the common industry standard, especially in flexible environments like cloud or software-defined networks.

But I will never block one to do this anyway.

What I fear: There are tons of various network devices out there. Every network device has a different set of configurations, a different set of features and a different API. Not every vendor is following standards here. It will be very hard to cover a lot of those devices. Even a good abstraction will not solve this (alone).

An alternative suggestion:
What do you think about an integration/support for Grafana / Prometheus instead of reinventing the wheel here?

As I said: For my/our use case we don't see a need for a special NOC Module like it is there right now. But I would be happy to merge improvements on this if one has the motivation to work on this.

andygrunwald:
@M4LuZ How do we continue with this PR?
Close it? We seem to have different approaches/opinions here (which is fine).

M4LuZ:
I remember that I wanted to write a longer reply to this, but it seems I never actually did that ;)
Gist of it was: 
Yes, there are a ton of tools that provide much more than the basic functionality this would provide.
But each additional tool would require a separate setup and interfacing.
I do not see that we save a lot of work while making LS a lot more complex to set up. 

BTW: the same argument can be made for almost any of the other modules. 
Board? Gallery? News? Tournaments? There are better standalone tools for all of that.
The strength of LS is not the offered functionality, but the integration of it 

Regarding this:
> There are tons of various network devices out there. Every network device has a different set of configurations, a different set of features and a different API.

Yup. But the basic functionality that I see so far for the NW-module is either standardised  or can be set by the user easily.
Either by having a SNMP OID mapping table or by providing the relevant switch console / shell commands with a placeholder.
And if that is not sufficient, then the interactive part should just be easily replaceable, so that interfacing can be changed if somebody really wants it.

I had a short moment where I thought to just merge and be done with it, but given that it does seem to be a functional module I would still would leave it in - until proven otherwise.

But happy to mark it for a rework, as I do see that it can be done better.

andygrunwald:
Fine for me so far.

PR #325, opened 2018-06-07
 Added symfony/debug component to get more information about errors

andygrunwald:
When you enable the `debugmode` in the configuration via

```
[lansuite]
debugmode = 1
```

PHP errors are shown this way:

standard-lansuite-db-error

The problem with this kind of errors: The root cause of Database errors (e.g. free calls) are not in the shown file and line. To get the real root cause, we typically need a stack trace.
The [symfony/debug](https://symfony.com/components/Debug) offers an easy way to print nice looking stack traces in error cases.
The same errors shown in the upper image are looking in the debug view like this:

symfony-debug

The old error handler is still in use, but only if the debugmode is disabled.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
 
         "smarty/smarty": "^3.1",
         "geshi/geshi": "^1.0.9",
-        "setasign/fpdf": "^1.8"
+        "setasign/fpdf": "^1.8",
+        "symfony/debug": "^3.4"
     },
     "require-dev": {
         "phpunit/phpunit": "^6.5",
--- a/index.php
+++ b/index.php
@@ -2,6 +2,8 @@
 // Composer autoloading
 require __DIR__ . '/vendor/autoload.php';
 
+use Symfony\Component\Debug\Debug;
+
 // Set error_reporting.
 // It is set to this value on purpose, because otherwise
 // LanSuite might not work properly anymore.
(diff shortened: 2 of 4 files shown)


index.php
```diff
@@ -2,6 +2,8 @@
 // Composer autoloading
 require __DIR__ . '/vendor/autoload.php';
 
+use Symfony\Component\Debug\Debug;
```
M4LuZ: Depending on comment above: Should only be referenced if debugging is on

composer.json
```diff
@@ -18,7 +18,8 @@
 
         "smarty/smarty": "^3.1",
         "geshi/geshi": "^1.0.9",
-        "setasign/fpdf": "^1.8"
+        "setasign/fpdf": "^1.8",
+        "symfony/debug": "^3.4"
```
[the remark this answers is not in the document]
andygrunwald: `require-dev` are dependencies that are not necessary to run LANSuite.
If an enduser enabled the debugmode (when the user downloads the released package), it is not there and a fatal error will be thrown.

In this case here, it don't add overhead.

index.php
andygrunwald: This is not possible, because `use` statements are only allowed on the top level and not inside scopes like `if` statements.

M4LuZ: oh, yeah, true that ;)

composer.json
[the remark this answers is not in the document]
M4LuZ: That is a weak argument, isn't it?
If you want to have a stack trace, then to fix a code / DB issue, thus you are developing.
But OK, mostly because we lack a better way to get end users to provide us with good error reporting

PR #357, opened 2018-09-09
#345: Fix manual translation query

### LINKED ISSUE #345: Manually changing translations not working
When changing translations in a module via the GUI in module-configuration the changes do not propegate.
Changing the translation manually in the translation.xml file for the module also does not make a difference.

## Expected Behavior

Changing a line, and hitting "editieren" should propegate the change in the line to system.

## Current Behavior

Changing a line, and hitting "editerien" does no change, the line is back to what it was before.

1. Enter translation tab on a module config page
2. Change a line in the translations
3. Click editieren
4. Go back to translation tab - Line will be back to standard.

## Your Environment

* Version used: Latest
* Operating System and version: Ubuntu Server 18.04 
* Docker Version: 18.06.1-CE
* Docker Compose Version: 1.22.0

M4LuZ:
which is rather a module in that context
Works now with this
fixes #345


andygrunwald:
Indeed. A good example why a good naming is important :)

PR #375, opened 2018-10-18
Update seatadmin.php

HtHNightwolf:
#373 
#374

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,6 +60,7 @@
             "LanSuite\\": "inc/Classes"
         },
         "files": [
+            "inc/base/define.php",
             "inc/Functions/T.php",
             "inc/Functions/FetchDataRow.php",
             "inc/Functions/FetchPostRow.php",
--- a/index.php
+++ b/index.php
@@ -125,9 +125,6 @@ function myErrorHandler($errno, $errstr, $errfile, $errline)
 
 $PHPErrors = '';
 
-// Read definition file
-include_once('inc/base/define.php');
-
 // Read Config and Definitionfiles
 // Load Basic Config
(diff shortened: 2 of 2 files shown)


andygrunwald:
Sadly, this will not fix the root cause of the `HTML_NEWLINE` constant.
The constant `HTML_NEWLINE` is used in many more places. See a code search here: https://github.com/lansuite/lansuite/search?q=HTML_NEWLINE&unscoped_q=HTML_NEWLINE

The constant `HTML_NEWLINE` is defined here: https://github.com/lansuite/lansuite/blob/master/inc/base/define.php
As you can see, it already has the value `<br/>` (like you used).

So the root cause seems to be a wrong handling of the constant in PHP-Strings. PHP is not able to replace constants in simple strings. The string has to "opened up".
So the solution to the first sentence would be instead of the original line

```php
array_push($questionarray, t('Sitzplatz für %1 reservierenHTML_NEWLINE(Ein evtl. zuvor für diesen Benutzer reservierter Platz wird freigegeben)', $new_user['username']));
```

we would need to use

```php
array_push($questionarray, t('Sitzplatz für %1 reservieren' . HTML_NEWLINE . '(Ein evtl. zuvor für diesen Benutzer reservierter Platz wird freigegeben)', $new_user['username']));
```

Have a detailed look at `' .` and `. '` around the `HTML_NEWLINE` constant.
To fix this for the codebase, this needs to be done for the whole codebase.

Let me check if I will have time for this today. If you are faster and motivated to do it, just write here that you are working on this (to avoid double work).

andygrunwald:
I've got some time for #378.
Please check it out there.

PR #379, opened 2018-10-20
Update composer dependencies

andygrunwald:
Some packages got updates in the last time.
This Pull Request updates all to the latest versions without raising major versions

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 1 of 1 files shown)


composer.lock
```diff
@@ -348,25 +348,28 @@
         },
         {
             "name": "myclabs/deep-copy",
-            "version": "1.7.0",
+            "version": "1.8.1",
```
M4LuZ: v.1.8.1 breaks compatibility with php 7.0, would suggest to go for a lower version instead

andygrunwald: True. Sorry, my bad.
I will cancel this PR and drop a new one

PR #382, opened 2018-12-15
#72 enforce unique email + username (configurable)

### LINKED ISSUE #72: Check required, if account was already registered with a given email adress
I just had a case where a user created three accounts with the same username and email address.
It seems that uniqueness of neither username nor email address is currently enforced.
But this should be the case for at least one of them. Which one is up for discussion, as both have their advantages and disadvantages

## Expected Behaviour
Entered user data should be checked against existing users and uniqueness of either username or email address should enforced. Both on application and database level.

## Current Behaviour
Users are able to create additional accounts with the same username and email address.
Those get a new userid assigned. It has not been tested what account is picked for login when the correct credentials are entered

## Possible Solution

Entered user data should be checked against existing users.
Uniqueness of either username or email address should be enforced. 
Preferably both on application and database level.

Create a new user, use same data as for one of the existing ones

## Context

This is needed to reduce the amount of dead accounts in the system and make it easier for both guests and administrators to find the right user account for a given username / email address

## Your Environment

* Version used: GIT:maluz/master
* Operating System and version: Win10
* Enabled features:

M4LuZ:
Enforce uniqueness of User email adresses on signon.
Also adds an option to do the same for usernames - activated by default.
This was already tested according to my own commit documentation, not sure any more why I didn't open the PR for this...

fixes #72


M4LuZ:
Ahh, this was already merged. Closing this here and removing the branch...

Issue #370, opened 2018-09-24
Smarty template processing throws warnings in PHP 7.2


M4LuZ:
## Expected Behavior

No warnings from the template anywhere

## Current Behavior

PHP throws warnings when smarty tries to process some template files, as it generates code that uses count() on non-countable objects, as described here:
https://www.smarty.net/forums/viewtopic.php?p=93082

## Possible Solution
Most propably this is fixed in a newer version of Smarty, so considering this a low-prio Ticket just to be considered when this is easy to fix

## Steps to Reproduce (for bugs)

1. Enable debugmode
2. Open e.g. user details
3. Enjoy the warning ;)

## Context

This causes issues when debugging one of the pages where e.g mastersearch is used.
So basically everywhere ;)
Only leaves the option to disable debugmode and continue without the viable information, or to manually reset the error reporting to E_ERROR only.
## Your Environment

* Version used: Git master
* Operating System and version: Win10
* Enabled features: ALL OF DEM!

mbraun81:
try to set sql mode without ONLY_FULL_GROUP_BY

aj-gh:
Should have been fixed in #386 .. close?

mrflobow:
Hi , is this still open and needs help ?
As I read replacing |count with |default:array()|count may fix the error.

andygrunwald:
I think we can close this issue.
LanSuite is on its way to being upgraded to PHP 8.0.
A lot of Pull Requests are made to fix several PHP Warnings.

It is still a long way to go, but we are progressing well.

This issue is not actionable with a particular error.

PR #386, opened 2019-01-24
mastersearch2: Fix PHP7.2 warning about count() on now non-countable objects

aj-gh:
PHP7.2 is more strict on what can be count()'ed and what not.
This first checks if the object is an array and only then does the counting; fixing #370 while at it.

This should fix the template "count(): Parameter must be an array or an object that implements Countable" error.


M4LuZ:
Seems that the code repo for PHP_CodeSniffer is unreachable atm. But code looks fine (to be tested)

Issue #224, opened 2018-04-29
Scheduling a release / TODOs for the next release


andygrunwald:
## Context

The last release of LanSuite was v4.2 from Mar 15, 2015.
Since then we did a couple of changes here and there.

What changes do we want/need to do before we ship a new release?

- [ ] Add a `CHANGELOG.md` with the listed changes
- [ ] Fix #72
- [ ] Fix #9 
- [ ] Fix #89
- [ ] Fix #145
- [ ] Fix #185
- [ ] Fix #222

What else comes into your mind @M4LuZ ?
What about #183 ?

I know that we talked about to test the new code at a party before we release it.
The thing is: Not so many people run this code at a party on master version (maybe you and me, but that's it).
And if we wait and keep changing, we pile up changes and changes and many people are "afraid" to upgrade.
So the coin has two sides.

What do you think?

M4LuZ:
I only see the following issues that we should really fix before release:
* #83 (to get away from unsalted MD5 hashes)
* #89 (to provide a complete package for any webspace-only installation)
* #214  (to provide detailed update documentation)
* And this one here: http://www.lansuite.orgapage.de/index.php?mod=bugtracker&bugid=1120&ms_page=0 as the PDF module is nonfunctional at the moment.
@byte666 seems to have fixed this already, but hasn't provided the code so far.

The other stuff has bothered nobody so far. 
Happy to include it also, but wouldn't consider it a dealbreaker

andygrunwald:
Okay great. I updated the checklist at the top of the issue. Lets work towards this :)

M4LuZ:
I would now also argue to include #90 (implement better upgrade handling) if we have any chance.

PR #387, opened 2019-01-24
Fix display of IP addresses in server module; fix binary IP handling in MasterForm

aj-gh:
Looks like this was broken since the database field was switched to varchar(16) for future IPv6 support in commit 16264ba71ca7174aeedef110e15bdb11dd84d4a9


M4LuZ:
Seems that the code repo for PHP_CodeSniffer is unreachable atm. But code looks fine (to be tested)

modules/server/search.inc.php
```diff
@@ -13,7 +13,7 @@
 $ms2->AddSelect('u.userid');
 $ms2->AddResultField(t('Name'), 's.caption');
 $ms2->AddResultField(t('Servertyp'), 's.type', 'ServerType');
-$ms2->AddResultField(t('IP-Adresse / Domain'), 's.ip');
+$ms2->AddResultField(t('IP-Adresse / Domain'), 'INET6_NTOA(s.ip) AS ip');
```
andygrunwald: Shouldn't be this `realip` then? Or doesn't it have a direct relation to the other usages?

aj-gh: Well... I initially fixed this code (overview page) and then noticed that the details view was still broken.
I didn't want to go far from "ip" here because that's how that field was always called.
However, in _show_details.php_  a select * is done which also returns "a.ip" as "ip". Now if I would have called the INET6_NTOA result "ip" as well that query would return both columns as "ip" ... which would become rather interesting I guess; at least confusing. I then looked at the amount of columns of the _server_ table and ended up with calling that field _realip_.
Reconsidering this, I guess the best way would be me changing this in _show_details.php_ from a SELECT * to specifying all the columns and also calling the resulting field "ip" there.
What do you think?

andygrunwald: > Reconsidering this, I guess the best way would be me changing this in show_details.php from a SELECT * to specifying all the columns and also calling the resulting field "ip" there.

Yes. I am for this. Personally, I like to avoid `SELECT *` and to make it explicit which fields you are requesting. Often this makes things easier. Do you want to adjust it in this PR or in a new one?

aj-gh: Will do it here.

aj-gh:
Not done yet; "Edit" is still broken.

aj-gh:
> Not done yet; "Edit" is still broken.
.. and fixed.

inc/Classes/MasterForm.php
```diff
@@ -506,7 +506,11 @@ public function SendForm($BaseURL, $table, $idname = '', $id = 0)
                     $db_query = '';
                     if ($this->SQLFields) {
                         foreach ($this->SQLFields as $val) {
-                            $db_query .= ", $val";
+                            if ($SQLFieldTypes[$val] == 'varbinary(16)' and $val == 'ip') {
```
andygrunwald: Why is a comparison needed by the type (`varbinary`)? I have the feeling that this will lead to a bug later on once we change this because the code relies on the data field.

aj-gh: Well, we already [have this in when adding/updating values](https://github.com/lansuite/lansuite/blob/36af57dbb00321b27895f98733f138be365d85ee/inc/Classes/MasterForm.php#L1122). This now fixes the same thing for retrieving values.
If we'd ever touch the fields we need to update both, however touching the fields would also mean we will most likely need to modify INET6_NTOA/ATON to something else anyway.
I think this is currently quite a clean approach because it applies these data modifications only if both field name and type match to what is required by these functions. Anything else and it is being dealt with like a regular database field without any modifications.
On future changes I'd assume that we'd either update both or add additional code for the necessary transformations which would also ensure that we'd stay compatible with older database layouts (e.g. the current one here).

aj-gh: Any news? I think this is currently one of the very last major open bugs of 5.0-dev/Master and users will certainly hit it.

andygrunwald: Thanks for providing more context. Fine with me.

andygrunwald: Related to @M4LuZ comment:
For the LanSuite v5, there might be more issues. Yep. I remember there was a ticket of things we aligned on. Not sure where it is right now.

Related to the `varbinary(16)` comparison. I am aware that this is not beautiful. But it is working and consistent across the code (as stated by @aj-gh). I think this is more important right now (when i see the status of this project).

andygrunwald: @M4LuZ I am waiting for your response/action before merging this. I am fine by merging this. Don't want to do it when you disagree.

aj-gh: > I'm sorry to disagree, there are still a good amount of issues left open, especially the upgrade handling from 4.2 -> 5.0
> That is also a place where the field conversion to `varbinary(16)` could be done to avoid this code here.

Even if you fully convert everything - you will still need either this or similar code unless you expect users to be able to parse IP addresses in varbinary format (as that is being shown to users when not doing INET6_NTOA on them, so without this code) ;)

aj-gh: > I'm sorry to disagree, there are still a good amount of issues left open, especially the upgrade handling from 4.2 -> 5.0

Well, missing upgrade handling doesn't make core features (which I consider the `server` module to be a part of) on a freshly installed 5.0-dev/Master unuseable - that's what I would consider a major bug. And of these there shouldn't be that many left (see below) given that we run 5.0-dev/Master for our upcoming event.

> For the LanSuite v5, there might be more issues. Yep. I remember there was a ticket of things we aligned on. Not sure where it is right now.

[Here is the ToDo list](https://github.com/lansuite/lansuite/issues/224) (that would probably need another update) while [this is a list of issues and PRs tagged for 5.0-RC](https://github.com/lansuite/lansuite/milestones/LanSuite%205.0%20RC) - seems manageable.

aj-gh: Push :)

M4LuZ: > Related to the `varbinary(16)` comparison. I am aware that this is not beautiful. But it is working and consistent across the code (as stated by @aj-gh). I think this is more important right now (when i see the status of this project).

^this, thus merging it while not being that happy about it ;)

PR #392, opened 2019-01-25
Don't try to create tables with NULL in primary keys..

### LINKED ISSUE #367: SQL Fail @ /index.php?mod=install&action=wizard&step=9
> Fehler:
> SQL-Failure. Database respondet: All parts of a PRIMARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead
>
>Query: 
>CREATE TABLE IF NOT EXISTS ls_ip_blacklist (ip varbinary(16) NULL , module char(20) NOT NULL default '' , PRIMARY KEY (ip) ) ENGINE = MyISAM CHARACTER SET utf8
>SQL-Failure. Database respondet: All parts of a PRIMARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead
>
>Query: 
>CREATE TABLE IF NOT EXISTS ls_ip_locklist (ip varbinary(16) NULL , module char(20) NOT NULL default '' , PRIMARY KEY (ip) ) ENGINE = MyISAM CHARACTER SET utf8

## Expected Behavior
Keine SQL-Fehler ;)
## Current Behavior
SQL-Fehler
## Possible Solution
All parts of a PRIMARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead
* Neue Subdomain erstellt für einen Test.
* Git Clone, Abhängigkeiten(Composer) und DB (+User) über Plesk Onyx installiert und angelegt.
* Installations Wizard gestartet.
* Nach Admin Erstellung bei der Modul-Auswahl oben auf weiter geklickt und wieder zur Admin Erstellung gelangt. Erneute Modul-Auswahl nun unten auf weiter geklickt -> Fehler Meldung
## Your Environment
PHP: 7.0.32-0ubuntu0.16.04.1 FPM/FastCGI
MySQL: 5.7.23-0ubuntu0.16.04.1

aj-gh:
.. as that will fail in any more recent MySQL/MariaDB versions or at least not work in the desired way. There are some SQL standard discussions about this, but anyway, we should not need any NULL values in an IP address field that is used for the specific purpose of IP matching, right?
See also: https://mariadb.com/kb/en/library/primary-keys-with-nullable-columns/
Should fix #367.


andygrunwald:
You are right. One question comes up: Should this be fixed in existing installations also?

aj-gh:
I don't think any handling for existing installations is required. Looking at the link I provided above it seems this will only fail either during a CREATE or when no value is provided for an INSERT, which shouldn't be the case given the structure of these tables.
I think we're good with fixing this only for new installs. Or are there any usecases where existing installs would recreate these tables without using the new (fixed) structure? Import maybe?

andygrunwald:
> Or are there any usecases where existing installs would recreate these tables without using the new (fixed) structure? Import maybe?

Not as far as I know. If this happens, we will provide an UPDATE/ALTER query. But for this we first would need a bug report / the information

PR #395, opened 2019-01-25
Fix 394 travis build

### LINKED ISSUE #394: No releases available for package "pear.php.net/PHP_CodeSniffer"
## Expected Behavior

Travis runs ;)

## Current Behavior

It seems that the Pear repo for PHP_CodeSniffer is dead

## Possible Solution

Switch to composer for it as require_dev dependency?

1. Commit something
2. See Travis log

## Context
Unable to merge anything into master

M4LuZ:
Obtains PHP_CodeSniffer by composer instead of the faulty Pear repo

fixes #394

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ cache:
     - $HOME/.composer/cache
 
 install:
-  - pear install PHP_CodeSniffer
   - phpenv rehash
   - composer install
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,54 +20,54 @@ install:
   - composer global require "squizlabs/php_codesniffer *"
   
 before_script:
-  - phpcs --version
+  - ~/.composer/vendor/bin/phpcs --version
 
 script:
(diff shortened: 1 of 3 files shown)


andygrunwald:
Oh, i saw this now. Will close #400 and will adjust this a little bit.

andygrunwald:
What I did and why:
I moved `squizlabs/php_codesniffer` to a development dependency. The reason is, it is not needed to run lansuite. By this, I changed the `composer install` to `composer install --dev`. So during a travis ci build, the dev dependencies are installed as well. This makes the `composer require global ...` obsolete, because `squizlabs/php_codesniffer` is installed like any other dependency. By this, it is more consistent and we have one way of installing dependencies. Another advantage is that you can run it locally.

I am merging this now, the build is green again.

M4LuZ:
Thanks for this.
Implementation is far superior to what I did to "get the damn thing working again" ;)

PR #403, opened 2019-01-26
Optimize php linting command to speed up CI builds on TravisCI

andygrunwald:
*Work in Progress*: This PR is work in progress and should not be merged right now. The basic idea/goal is to speed up the CI build times. 

Now we run/What it does:
- PHP Linting: 8 checks in parallel
- PHP Linting: Only output PHP errors, if a file is valid it won't be outputted anymore
- Bundle phpcs calls to reduce startup time


M4LuZ:
@andygrunwald: Is this still WIP or ready to be merged?

andygrunwald:
Still work in progress.
Until now i didn't made big improvements here.

M4LuZ:
> Until now i didn't made big improvements here.

I think we will also not see any improvements here as the resources granted per instance are limited and I think that parallel execution may actually slow things down a bit.
What may make sense instead is to restrict linting via `phpcs` to the set of changed files as these are the only ones that contain changes to be checked
Syntax checks would need to run at least once on all files (due to potential dependencies in non-modified files) before merging, but may also be reduced to just the changed files for immediate checks.

No clue how to implement that though ;)

M4LuZ:
@andygrunwald please have a look here.
Was able to achieve a large speedup by just selecting the modified files.
We do not need to check existing files, as they are per definition OK if they are in the master.
I thought that it still makes sense that the php syntax runs on every file but it appears that no dependencies are checked, thus also here only changes need to be validated.

I ran a few test cases in the branch 'test-travis-speedup' and noted the following:

**git diff does not return added files with the current call, so that still needs fixing here.**
I'm not that confident with neither Travis nor Git, so would need your help here!

M4LuZ:
OK, I give up.
Thought that using `$TRAVIS_COMMIT`would help, but Travis creates a merge commit and that is what is contained in that variable. But somehow I'm unable to obtain the changeset from there.
Help pl0x!

M4LuZ:
File selection seems OK (new and modified files checked, deleted files ignored)
But the verification now fails to fail, seems the return code is not validated correctly

M4LuZ:
yeah, about that. A grep has to be added to the first call to make sure that only php files are passed for `php -l`

M4LuZ:
OK, seems we are now good on all points.
There is still a minor cosmetic issue as `grep "\.php$"` exits with an error if there is no change to a php file, but directly the next line forces the job to be OK for exactly that case.

@andygrunwald would be happy if you give it a spin and add your two cents here ;)

andygrunwald:
From a quick look, it looks good. Thanks for putting the hard work in it. I hope this speeds up the process a lot :)

One side option:
Once we modify a big file that is not fully PSR-1/PSR-2 compliant, the check PHP_CodeSniffer check will fail. The reason is, that there are no more "exclusions" anymore like before. Before we applied several exclusions on modules and more.

This will be maybe hard if someone touches a big file.
I am personally fine with it. Many modern IDEs have an "auto format feature" for PSR-1 or PSR-2.

IMO: Go ahead, merge it.

M4LuZ:
> Once we modify a big file that is not fully PSR-1/PSR-2 compliant, the check PHP_CodeSniffer check will fail. 

Yes, but fixing that should be fast and most of the files should be OK already.
We may want to exclude the line length again, but that can be handled separately.

> I hope this speeds up the process a lot :)

It does, see the CI tests in the branch [`test-travis-speedup`](https://github.com/lansuite/lansuite/compare/test-travis-speedup) 
Basically this now takes the VM boot time (around a minute) + a few seconds.
So there is not much left to optimize besides caching the composer downloads somehow.

But merging this for now

BenniGotchi:
> We may want to exclude the line length again, but that can be handled separately.

Since I'm struggling with this rule atm here: #441 
And I think there are pros and cons for this check, I created a seperate Issue to discuss this rule in more depht: #442

M4LuZ:
raised PR #444 to re-exclude line-length

PR #428, opened 2019-02-03
Partylist: Module improvements

### LINKED ISSUE #302: Replace self-developed HTTP client in partylist/GetSite.php by native functions
## Expected Behavior

Solution to get Party infos from other pages should support redirects and HTTPS.

## Current Behavior

Currently it does not, because it uses a self-built HTTP 1.0 GET request

## Possible Solution

Either use native PHP functionality or replace this by a simple interface

Add a new party in the partylist module. Add URL to any other LS-installation with redirect or HTTPS

aj-gh:
Switch to native PHP functions for file retrieval and support HTTPS party URLs.
Fixes #302.

### DIFF
--- a/modules/partylist/Functions/AddSignonStatus.php
+++ b/modules/partylist/Functions/AddSignonStatus.php
@@ -7,19 +7,19 @@
  */
 function AddSignonStatus($lsurl, $show_history = 0)
 {
-    global $xml, $dsp, $HTTPHeader, $func;
+    global $xml, $dsp, $func;
 
     if (substr($lsurl, strlen($lsurl) - 1, 1) != '/') {
         $lsurl .= '/';
     }
-    if (substr($lsurl, 0, 7) != 'http://') {
+    if (substr($lsurl, 0, 7) != 'http://' && substr($lsurl, 0, 8) != 'https://') {
         $lsurl = 'http://'. $lsurl;
     }
     $lsurl .= 'ext_inc/party_infos/infos.xml';
     $content = GetSite($lsurl);
 
     if (!$content) {
-        return '<div class="infolink" style="display:inline">'. t('infos.xml fehlt') .'<span class="infobox">'. $lsurl .HTML_NEWLINE.HTML_NEWLINE. str_replace("'", "\\'", str_replace('"', "'", str_replace("\r\n", HTML_NEWLINE, $HTTPHeader))) .'</span></div>';
+        return '<div class="infolink" style="display:inline">'. t('infos.xml fehlt') .'<span class="infobox">'. $lsurl .'</span></div>';
     } else {
         $system = $xml->get_tag_content_array('system', $content);
         // Version 3.0 XML-File
--- a/modules/partylist/Functions/AddSignonStatus.php
+++ b/modules/partylist/Functions/AddSignonStatus.php
@@ -19,7 +19,14 @@ function AddSignonStatus($lsurl, $show_history = 0)
     $content = GetSite($lsurl);
 
     if (!$content) {
-        return '<div class="infolink" style="display:inline">'. t('infos.xml fehlt') .'<span class="infobox">'. $lsurl .'</span></div>';
+        $content_error = '';
+        if (!extension_loaded('openssl') && strpos($lsurl, 'https://')!==false) {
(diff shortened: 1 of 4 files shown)


modules/partylist/Functions/GetSite.php
```diff
@@ -3,51 +3,8 @@
...
-        $HTTPHeader = substr($cont, 0, strpos($cont, "\r\n\r\n"));
-
-        $StatusCode = substr($HTTPHeader, strpos($HTTPHeader, ' ') + 1, 3);
-        if ($StatusCode != 200) {
-            return '';
-        }
-
-        return substr($cont, strpos($cont, "\r\n\r\n") + 4);
-    }
+    return @file_get_contents($url, false, stream_context_create(array('http' => array('timeout' => 10))));
```
M4LuZ: This AFAIK only works for HTTPS if OpenSSL is enabled as module. 
May thus fail silently when any site is only reachable via HTTPS.

modules/partylist/show.php
```diff
@@ -8,7 +8,7 @@
         if (substr($row['ls_url'], strlen($row['ls_url']) - 1, 1) != '/') {
             $row['ls_url'] .= '/';
         }
-        if (substr($row['ls_url'], 0, 7) != 'http://') {
+        if (substr($row['ls_url'], 0, 7) != 'http://' && substr($row['ls_url'], 0, 8) != 'https://') {
             $row['ls_url'] = 'http://'. $row['ls_url'];
```
M4LuZ: As mentioned before. Should default to HTTPS if available

modules/partylist/show.php
```diff
@@ -62,7 +62,7 @@
         p.partyid = %int%", $_GET['partyid']);
     $framework->AddToPageTitle($row["name"]);
 
-    if (substr($row['url'], 0, 7) != 'http://') {
+    if (substr($row['url'], 0, 7) != 'http://' && substr($row['url'], 0, 8) != 'https://') {
         $row['url'] = 'http://'. $row['url'];
```
M4LuZ: Same here

modules/partylist/Functions/AddSignonStatus.php
```diff
@@ -7,19 +7,19 @@
...
 {
-    global $xml, $dsp, $HTTPHeader, $func;
+    global $xml, $dsp, $func;
 
     if (substr($lsurl, strlen($lsurl) - 1, 1) != '/') {
         $lsurl .= '/';
     }
-    if (substr($lsurl, 0, 7) != 'http://') {
+    if (substr($lsurl, 0, 7) != 'http://' && substr($lsurl, 0, 8) != 'https://') {
         $lsurl = 'http://'. $lsurl;
```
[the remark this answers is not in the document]
M4LuZ: ...if available, of course!

aj-gh:
Defaulting to HTTPS:
Browsers don't default to HTTPS yet. I would be all for it however that is not the case yet so I don't see the point in enforcing it for remote LanSuite URLs if users are used to see different behaviour in their browsers.
Checking if HTTPS is available would require us doing an extra callout and even that doesn't ensure we get the right content. I would prefer to keep adding "http://" as-is if nothing was provided by the user. If the remote LanSuite URL is https enabled the user can add the link with a prefix of "https://" and that would just work (at least after this PR).
In my opinion - when common browsers start defaulting to https:// it certainly makes sense to modify these defaults, before that I wouldn't just do it yet.
Changing this might also break existing sites that use `partylist` functionality without providing a protocol prefix and rely on it defaulting to http://

OpenSSL availability:
You are right. However this is not the only possibility on why this could silently fail - [allow-url-fopen](http://de2.php.net/manual/en/filesystem.configuration.php#ini.allow-url-fopen) being the other possible cause for issues.
As the `Discord` module relies on the same code I guess we should include a check for OpenSSL and allow-url-fopen in the installation/system compatibility check code with the appropriate warnings. What do you think?
In any case I'll look into adjusting both this PR and `Discord` to check for both because the admin could have ignored the warnings and users would then wonder why it doesn't work as expected.

M4LuZ:
Argumentation is fair and while I would like to see us moving towards https-by-default, I'm fine with passing this through for now and having another go at some point in the future

Issue #384, opened 2019-01-22
Edit User as Admin+


makno:
## Expected Behavior

Changing user settings i.e. type user to admin should work flawless.

Currently,  'CheckValidUsername' and 'CheckValidEmail' as well as the second Email field to check valid mails, don't work.

## Current Behavior

Errors prevent from saving are 'User already exists', 'Email already exists' and that the second email is not equal to the first entry. (Because 2nd is empty by default)

## Possible Solution

$mf->AddField(t('Benutzername '.$auth['type']), 'username', '', '', $optional); // MAKNO removed last argument: ,'CheckValidUsername'
$mf->AddField(t('E-Mail'), 'email', '', '', '');  // MAKNO removed last argument: , 'CheckValidEmail'
// MAKNO because it doesn't work:  $mf->AddField(t('E-Mail wiederholen'), 'email2', '', '', '');

## Your Environment

* Version used: 5.0-dev
* Operating System and version: Debian
* Enabled features: Default

aj-gh:
I think the whole email- and username verification needs a review.
ea88add6fbf5775586887c80e95eea7a2bd999e8 introduces the email2 field. By a quick look this already seems to break the newmail function of the mail module when it's enabled for unauthenticated users as that uses the same CheckValidEmail function:https://github.com/lansuite/lansuite/blob/db4b61f3d97dae0069476a061c2acacc639a2cd5/modules/mail/newmail.php#L116
For usrmgr - if an admin modifies an user email2 won't be set (unless the admin is really bored), so this should be bypassed for edits as long as the email address itself is not being modified.
#368 introduces checks for duplicate usernames and email addresses which would theoretically also affect mail/newmail (as above). If I'm not mistaken this also introduces issues for admin edits as the username always exists unless it was just modified in that call by the admin.

Not sure what the best way forward would be.

I guess the duplicate-checks should be moved into separate validation routines if possible and only trigger when they make sense. The email verification field (email2) could probably be validated in MasterForm similar to passwords where this is triggered via a "IS_NEW_PASSWORD" flag - this would also remove the rather ugly access to $_POST in the generic validation function.

Ideally the goal would be that both validations and duplicate-checks are still in place for new user registrations and for modifications if that particular field is being modified, restoring both user modification functionality and mail functions for non-authenticated users (if enabled).

PR #429, opened 2019-02-03
Usrmgr: Allow identical username when username was not changed

aj-gh:
Fixes one of the issues mentioned in #384 where admin users are no longer able to modify user data as the username (obviously) already exists. This makes that check only trigger when the username is not identical to the one of the user currently being modified.

### DIFF
--- a/modules/usrmgr/Functions/CheckValidUsername.php
+++ b/modules/usrmgr/Functions/CheckValidUsername.php
@@ -9,7 +9,11 @@ function CheckValidUsername($username)
 {
     global $cfg, $db;
     if ($cfg['signon_username_unique']) {
-        $row = $db->qry_first('SELECT * FROM %prefix%user WHERE username=%string%', $username);
+        if (isset($_GET['userid'])) {
+            $row = $db->qry_first('SELECT * FROM %prefix%user WHERE username=%string% AND userid!=%int%', $username, $_GET['userid']);
+        } else {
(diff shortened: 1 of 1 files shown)


modules/usrmgr/Functions/CheckValidUsername.php
```diff
@@ -9,7 +9,11 @@ function CheckValidUsername($username)
 {
     global $cfg, $db;
     if ($cfg['signon_username_unique']) {
-        $row = $db->qry_first('SELECT * FROM %prefix%user WHERE username=%string%', $username);
+        if (isset($_GET['userid'])) {
```
M4LuZ: Hmmm.
I would rather use $auth here to get the current user details (if he is editing his own account) and only use $_GET if we have a valid admin session (and thus are editing a different account then what is in $auth).
Your thoughts?

aj-gh: As users [are not allowed to modify their usernames themselves](https://github.com/lansuite/lansuite/blob/33e5b3b6358e06fc37246c3539ae99e94f4f15ab/modules/usrmgr/add.php#L21-L25) I don't think that is an issue. It would be if that restriction won't be there though.
The `usrmgr` code also has checks to validate that `$auth['userid']` equals `$_GET['userid']` so providing a random userid as non-admin shouldn't work either.
I can modify this PR and change that code or we go with it as-is.

M4LuZ: No, that is fine by me then ;)

Issue #415, opened 2019-01-29
Translation loading broken for generic/global translations


aj-gh:
## Expected Behavior

Translations from generic/global translation files such as `inc/language/System_translation.xml` or `inc/language/DB_translation.xml` should be properly read, parsed and used.

## Current Behavior

It seems these files are only read when the translation-mode is "xml" before the database is configured. Afterwards they seem to be no longer read and used.
Also if they are read it seems they end up in a cache but are not called because the cache is built around the module name which doesn't seem to match on these generic/global translations (but I didn't investigate this further).

## Possible Solution

Translation code needs some rework.
See also #344.

## Steps to Reproduce (for bugs)

Switch language to anything but German and notice that even the most basic translations contained in these generic files are not applied.

## Context

My assumption would be that the System/DB translations are supposed to fill in for very common phrases that shouldn't be put into each modules separate translation file. Well, this currently doesn't seem to work that way.

## Your Environment

* Version used: 5.0-dev / Master
* Operating System and version: Debian 9 / Docker from Master

M4LuZ:
Fixed with merge of #650

PR #436, opened 2019-02-19
WIP: Initial implementation of caching by Symphony/cache based on PSR…

### LINKED ISSUE #432: Implement general caching mechanism
## Expected Behavior

LS should provide a general cache functionality to store a set of data with a timeout.
This could use a local file, a db table or APCu//memcache when enabled.

## Current Behavior

Separate implementations everywhere.
* Discord (see #422)
* TS3 (see #292)
* Smarty (although not sure if we can integrate this)
* Maps - potentially (see #301)
* Partyinfo (see #428)

## Possible Solution

implement a new core class for this, that automatically determines the best storage system and then provides an interface to get, store, update and remove datasets based on a key.
Could also include a callback function to refresh data content after timeout has elapsed to make autorefresh on a schedule possible

M4LuZ:
**This is still work in progress.** 
Had a few minutes on the road and started with cache initialisation and caching of the config file.
Already opening the PR to get feedback on the implementation.
This is intended as basis for the rework of the translation module, as that will benefit heavily from caching.

fixes #432


composer.json
```diff
@@ -20,7 +20,8 @@
         "geshi/geshi": "^1.0.9",
         "setasign/fpdf": "^1.8",
         "paypal/rest-api-sdk-php": "^1.13",
-        "symfony/debug": "^3.4"
+        "symfony/debug": "^3.4",
+        "symfony/cache": "* v4.2.3"
```
andygrunwald: I am not sure what `"* v4.2.3"` means.
`*` means every version. `v4.2.3` means version 4.2.3.
I would be in favor of `^4.2` or `v4.2.3` (explicit defining and set it fixed to v4.2)

index.php
```diff
@@ -125,30 +126,44 @@ function myErrorHandler($errno, $errstr, $errfile, $errline)
 
 $PHPErrors = '';
 
-// Read Config and Definitionfiles
-// Load Basic Config
-if (file_exists('inc/base/config.php')) {
-    $config = parse_ini_file('inc/base/config.php', 1);
+// Initialize Cache. Go for APCu first, filebased otherwise. DB adaptor to be used when we implement PDO.
```
andygrunwald: `DB adaptor` => `DB adapter`

modules/discord/Classes/Discord.php
```diff
@@ -46,25 +46,18 @@ public function __construct($discordServerId = ''){
      */
     
     public function fetchServerData(){
-        global $cfg;
+        global $cfg, $cache;
```
andygrunwald: Can we use a proper Dependency injection here?
global variables make it harder and harder. I know that we have them all over the place right now, but while introducing a new system we can make it better.

modules/install/wizard.php
```diff
@@ -26,7 +26,7 @@
             $row = $db->qry_first("SELECT email FROM %prefix%user WHERE email=%string%", $_POST["email"]);
 
             // If found, update password
-            if ($row['email']) {
+            if ($row !== false) {
```
andygrunwald: What is wrong here with the `if ($row['email']) {` comparision?

composer.json
M4LuZ: Beats me, that is what Netbeans auto-inserted. Happy to change that.

index.php
M4LuZ: "adaptor" is correct British English (but also "adapter")

modules/discord/Classes/Discord.php
M4LuZ: I do see that dependency injection has its uses and also that global variables should be avoided unless there is a good reason for them.
But: I would argue that this is case where they make sense.
This is a static, non-changing object reference that is used everywhere, so I do see the benefit of one global instance.

> global variables make it harder and harder.

What exactly?
I would hate to see functions where the first five or six parameters are only global dependencies.
Also: Refactoring is quite simple with globals. Just search for the variable name....

In case of an individual set of data/object, it should be passed as argument, complete agreement there.

modules/install/wizard.php
M4LuZ: Because `qry_first` does not return an array if no row is found and thus this would cause a warning, as it tries to access an array element of an boolean.

modules/discord/Classes/Discord.php
Dwarfex: If Dependency Injection was used in combination with the [PSR Simple Cache Interface (](https://packagist.org/packages/psr/simple-cache) Modules could decide on their own if they prefer f.e. file, database or in memory cacheing.

M4LuZ: > If Dependency Injection was used in combination with the [PSR Simple Cache Interface ](https://packagist.org/packages/psr/simple-cache) Modules could decide on their own if they prefer f.e. file, database or in memory caching.

As the cache implementation is done in each module separately that is possible in any case.
Also with DI it wouldn't be the module that decides what cache is used, but the caller...

If a module needs a separate cache pool implementation, then that is possible as the required sources are included with this PR forward.
But if just a simple cache is required, then this code provides that.

composer.json
M4LuZ: fixed as suggesteed

M4LuZ:
push @andygrunwald 
Would appreciate to get this in short-term, as the also almost finished rework of the translation module is based on this

andygrunwald:
LGTM. Iterations can be done at every time anyway.

M4LuZ:
Ran some tests on my laptp with jmeter.

### 20 threads, a request every 500 ms to load the index page, 10 repeats, no ramp-up time

Variant |  Avg. response | Min response | Max response
------------ | ------------- | ------------- | -------------
Old system | 429 ms | 119 ms | 1420 ms
Translation File cache | 465 ms | 127 ms | 952 ms 
Translation APCu cache | 409 ms | 107 ms | 1013 ms

### 200 threads, a request every 500 ms to load the index page, 10 repeats, no ramp-up time

Variant |  Avg. response | Min response | Max response
------------ | ------------- | ------------- | -------------
Old system | 8666 ms | 235 ms | 13343 ms
Translation File cache | 8895 ms | 394 ms | 12447 ms 
Translation APCu cache | 8957 ms | 314 ms | 14404 ms

The results seem strange so far, 
I expected a noticeable impact at least in the high-load case.
Performance seems to be CPU-limited, as it stayed at 100% during the load tests and the I/O-related improvements here do not make a noticeable difference.

Seems that more profiling is needed in general to see where we have performance bottlenecks

M4LuZ:
I leave this open for a bit for everyone to test and will merge this in about a week then
Rework of the translation functionality (#415) and the partymap caching (#438) are/will be based on this

Issue #440, opened 2019-03-22
Fatal Error when trying to import .xml Dumps while Installation


BenniGotchi:
## Expected Behavior

While Installation of Lansuite: Upload .xml dump of LansuiteData.
Import works flawless

## Current Behavior

Import fails with a Fatal error:
```
Fatal error: Uncaught Error: Call to a member function GetUploadFileType() on null in D:\MyProjects\git\lansuite\modules\install\wizard.php:239 Stack trace: #0 D:\MyProjects\git\lansuite\index_module.inc.php(55): include_once() #1 D:\MyProjects\git\lansuite\index.php(437): include_once('D:\\MyProjects\\g...') #2 {main} thrown in D:\MyProjects\git\lansuite\modules\install\wizard.php on line 239
```

## Possible Solution

Fix name inconsitency in 
https://github.com/lansuite/lansuite/blob/b75af3c28d17596cad868e2601bb32bb364a1e7a/modules/install/wizard.php#L7
https://github.com/lansuite/lansuite/blob/b75af3c28d17596cad868e2601bb32bb364a1e7a/modules/install/wizard.php#L239

## Steps to Reproduce (for bugs)

1. Perform a fresh installation with current Master
2. Select 'Import XML' and provide a Lansuite xml dump

## Context

Problem only present in the install wizard.
After a 'clean' installation, without importing old data, an afterwards performed import via the Admin Panel works.

## Your Environment

* Version used: current master
* Operating System and version: Windows 10 64bit, XAMPP, PHP7.3.2, MariaDB Version 10.1.38

M4LuZ:
I guess this can be closed now with the merge of #441?

BenniGotchi:
Yes.
How can that be handled in an automated way?
With #442 and #444 the issie was closed automatically with finishing the PR?

M4LuZ:
> With #442 and #444 the issie was closed automatically with finishing the PR?

The statement `fixes #442` in the pull request does all the magic.
That automatically closes the referenced issue as the PR is supposed to fix it
Use `refs #442`  if you just want to create a reference in the related ticket

BenniGotchi:
OK, thanks for the explanation 👍

Am Di., 9. Apr. 2019 um 14:33 Uhr schrieb M4LuZ <notifications@github.com>:

PR #441, opened 2019-03-22
Update wizard.php

BenniGotchi:
Fix fatal error in Install wizard, when trying to import Lansuite XML dump

Inconsitency in variable name removed

refs #440

### DIFF
--- a/modules/install/wizard.php
+++ b/modules/install/wizard.php
@@ -4,8 +4,8 @@ if ($_POST["resetdb"]) {
     $db->success = 0;
 }
 $importXml = new \LanSuite\XML();
-$installImport = new \LanSuite\Module\Install\Import($importXml);
-$install = new \LanSuite\Module\Install\Install($installImport);
+$import = new \LanSuite\Module\Install\Import($importXml);
+$install = new \LanSuite\Module\Install\Install($import);
 
 // XML is a global requirement during installation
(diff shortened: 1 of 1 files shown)


BenniGotchi:
@M4LuZ, @andygrunwald : Any Idea, why the automatic checks fail?
There are some warnings about line lenght:

    23 | WARNING | Line exceeds 120 characters; contains 137
     |         | characters (Generic.Files.LineLength.TooLong)

Is this causing the fail?

andygrunwald:
@BenniGotchi The check fails because of this recent change: https://github.com/lansuite/lansuite/pull/403#issuecomment-473046255

modules/install/wizard.php
```diff
@@ -4,8 +4,8 @@
     $db->success = 0;
 }
 $importXml = new \LanSuite\XML();
-$installImport = new \LanSuite\Module\Install\Import($importXml);
-$install = new \LanSuite\Module\Install\Install($installImport);
```
andygrunwald: Maybe I don't see it, but there is only the change of the variable name, right?
What does the Fatal error say? Or why will another name fix it?

andygrunwald: Ok. Saw it here. Sorry: https://github.com/lansuite/lansuite/issues/440

PR #444, opened 2019-04-05
Re-added exclusion of line-length checks

### LINKED ISSUE #442: Check for line length limit (PSR-2) - how to handle violations?
## Expected Behavior

To be discussed here in this issue ;)

## Current Behavior

With implementation of #403 Code Style checks for PSR-1 and PSR-2 are applied to all modified php files only, but the former existig exceptions were dropped.

Idea is, that unchanged files are still allowed to hold some CodeStyle exceptions, but all files that are touched, should be modified according to the current code styling rules.
In principle a good approach, I would say. 

But there is now a check active, where I would like to see a discussion for: 
The PSR-2 states for the line length:

    The soft limit on line length MUST be 120 characters; 
    automated style checkers **MUST warn but MUST NOT error** at the soft limit.

This is also, what the automatic checks do - they raise a Warning for these violations.
But the Travis-CI job returns a fail because of these warnings.

In #403 @andygrunwald mentioned, that in almost every IDE there are auto formatting features present, that can help for getting the coding standard fulfilled.
I have some doubts, that they will help for the line length.
 * All the code was put through automated code reformatting tools already, but they obviously didn't take care about (all) long lines
 * I tried the 'Format' feature in the  Eclipse PDT - didn't help.

So I think keeping up this strict handling for the line length would mean a lot of manual editing for the people, who want to contribute.
Worst case they get bored, because they want to contribute a small change but would need to fix a lot of other stuff not really related to the original topic...

## Possible Solution

accept warning for the line length in the automatic checks ?

Propose IDE setups, where the Code style rules are already checked before committing changes.
This would make it much easier, to fix deviations 'on the fly' ?

M4LuZ:
Ignore line-length again

fixes #442

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,5 +31,5 @@ script:
   - CHANGED_PHP_FILES=`echo "$CHANGED_FILES" | grep "\.php$"`
   - if [ -z "$CHANGED_PHP_FILES" ]; then travis_terminate 0; fi; 
   - echo "$CHANGED_PHP_FILES" | xargs -n1 -P4 php -l -n | (! grep -v "No syntax errors detected" )
-  - echo "$CHANGED_PHP_FILES" | xargs -n1 -P4 bin/phpcs --standard=PSR1,PSR2 --extensions=php -s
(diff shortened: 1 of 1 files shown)


BenniGotchi:
@andygrunwald please also consider my comment in #442 
thx

PR #455, opened 2020-10-20
Update Dockerfile

t-h-e:
upgrade base image to php 7.4.11
as well as necessary dependencies
to be able to create a docker image for lansuite again

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.0-fpm
+FROM php:7.2.34-fpm
 
 COPY . /code
 
@@ -10,6 +10,7 @@ RUN apt-get update \
         libsnmp-dev \
         snmp \
         unzip \
+        libzip-dev \
     && docker-php-ext-install -j$(nproc) mysqli snmp \
     && docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ \
     && docker-php-ext-install -j$(nproc) gd \


andygrunwald:
Hey @t-h-e,
thanks a lot for this pull request.
The changes are looking good and I would love to merge this. However, jumping from PHP 7.0 to PHP 7.4 might not be as straightforward.
Especially in LanSuite, this source is quite old and I am not 100% sure if it runs as expected on PHP 7.4.
That is the reason why I am not hitting the merge button immediately.

@M4LuZ What do you think?

t-h-e:
No worries. I should have mentioned that I have not tested it further than just checking if the lansuit setup starts.
I'm trying to migrate from a rather old lansuite version with a few minor changes to the current version. While also migrating the changes. If the changes still seem useful, I will create pull requests for them.
Anyway I will have to test everything after migrating anyway, so for me it's also fine to leave the pull request here until further confirmation that lansuite works properly with php 7.4
At least here the change is already accessible to everyone 😃

M4LuZ:
I would propose to change dockerfile to PHP 7.2 at first (as that is fairly certain to work) and upgrade travis to build against 7.2 -7.4.
If we don't get any complaints from there we should give 7.4 a shot and update version then

t-h-e:
7.2 is fine for me as well for now, but keep in mind that 7.2 is nearly end of life as well.
see https://www.php.net/supported-versions.php

M4LuZ:
That is well known and understood. 
Reason why I specifically mention PHP 7.2 is because that is the highest version we test against in Travis and thus had at least some test coverage by the CI-System. 
I agree that we should move on to 7.4 ASAP and only see that as stop-gap measure till then.

M4LuZ (approved):

Issue #456, opened 2020-11-10
MySQL/MariaDB sql_mode default STRICT_TRANS_TABLES 


t-h-e:
`STRICT_TRANS_TABLES` is part of the defaults in mariadb and mysql
https://mariadb.com/kb/en/sql-mode/#defaults
https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sql-mode-strict

Which is a good thing, as it helps find errors.

The issue is that lansuite heavily relies on MySQL/MariaDB to implicitly convert values and other "magic" (e.g. truncating values, setting default values although no default value was specified).

e.g. lansuite often tries to set an empty string independent of the column type (`field = ''`). Depending on the datatype it can be converted to `0`, `enum(0)`, `false` or `0000-00-00 00:00:00`.

lansuite should be able to work with the default sql_mode values of MySQL/MariaDB. Therefore some SQLs need to be adapted to set values explicitly and default values need to be defined for some columns.

This issue may be used obviously for discussion, but also as reference for all changes that are submitted regarding `STRICT_TRANS_TABLES`

t-h-e:
I have created a few pull request and would highly appreciate feedback (@andygrunwald, @M4LuZ)
These changes may not fix all issues, but also should not impair using older versions of MySQL/MariaDB

PR #457, opened 2020-11-10
Set all NOT NULL fields when adding Admin during wizard

t-h-e:
fixes parts of https://github.com/lansuite/lansuite/issues/456

### DIFF
--- a/modules/install/wizard.php
+++ b/modules/install/wizard.php
@@ -35,7 +35,7 @@ switch ($_GET["step"]) {
             // If not found, insert
             } else {
                 $db->qry(
-                    "INSERT INTO %prefix%user SET username = 'ADMIN', firstname = 'ADMIN', name = 'ADMIN', email=%string%, password = %string%, type = '3'",
+                    "INSERT INTO %prefix%user SET username = 'ADMIN', firstname = 'ADMIN', name = 'ADMIN', email=%string%, password = %string%, type = '3', lastlogin = NOW(), comment = '', birthday = '0000-00-00', signature = ''",
                     $_POST["email"],
                     md5($_POST["password"])
                 );


modules/install/wizard.php
```diff
@@ -35,7 +35,7 @@
             // If not found, insert
             } else {
                 $db->qry(
-                    "INSERT INTO %prefix%user SET username = 'ADMIN', firstname = 'ADMIN', name = 'ADMIN', email=%string%, password = %string%, type = '3'",
+                    "INSERT INTO %prefix%user SET username = 'ADMIN', firstname = 'ADMIN', name = 'ADMIN', email=%string%, password = %string%, type = '3', lastlogin = NOW(), comment = '', birthday = '0000-00-00', signature = ''",
```
andygrunwald, line 38: What do you think about setting the schema default to those values?
This would avoid errors once we insert the data in different places in the applications?

t-h-e: I thought about it and decided against it in this case for 2 reasons.

1. You do not want to have defaults for every field. In some cases it is on purpose to not have a default to avoid weird special cases that you then have to take care of, when you read from the database.
2. This is the wizard and it is kind of different from the rest of the application. For creating the admin you do not have the whole form to create the user (we could change that of course, but currently it is easy and quick to setup lansuite, which I like). I don't think we should create defaults just for the wizard.

So far, it seems fine for the rest of the application. Also, lastlogin can not be handled by a default, as NOW() changes. You would need a trigger for that. Unless there is a really good reason for a trigger, I would like to avoid it, as it usually leads to some kind of confusion why values change when writing to the database.

M4LuZ: What about the other option of changing field definition to `NULL` from `NOT NULL` where it is OK and intented to leave fields empty.
As I guess that that is cleaner than to try in semi nonsensical values just to satisfy a `NOT NULL` requirement that does not need to be there in the first place.
What do you think?

t-h-e: In general I agree. If a field does not need to be filled it should be kept empty. There is no need to set weird values.
The issue is that lansuite/MySQL up to now did exactly that. Just switching to `NULL` values instead has a high risk of breaking other parts of the application.
This being just the wizard that creates the admin user, I don't think it is worth the risk setting `NULL`.

M4LuZ: >  Just switching to `NULL` values instead has a high risk of breaking other parts of the application.

Not seeing how removal of a constraint can cause an issue. 
I was thinking that maybe leaving fields empty after fields are set to `NULL` could lead to an issue, but then remembered that this the whole point of the PR ;)
So would be very interested to understand where you perceive a risk, as it appears that I have a blind spot here.

t-h-e: I'm not sure I follow.

*Situation before this PR:*
Some columns cannot be set to `NULL`. In the past MySQL, in case of a `NULL` value just assumed which value to set. There was no `NULL` value in the database or in the application.

*Situation after the PR:*
Some columns cannot be set to `NULL`. Nowadays MySQL will throw an exception in case of trying to set a `NULL` value in a column that has a `NOT NULL` constraint. Therefore, this PR sets the values the were set in the past by MySQL anyway. So again, no `NULL` values in the database or in the application.

*New suggestion:*
Changing some columns from `NOT NULL` to `NULL` instead of changing the SQL. Then the database will contain `NULL` values. No issue with the database. But the application did never have any `NULL` values before. So we would need to check all locations where these columns are used and check if `null` is correctly handled.
e.g.: inseat of `if ($column === $defaultValue)`, we would need to check `if (is_null($column))`

If I misinterpreted the suggestion, let me know.

M4LuZ (approved):
Sorry for the _very_ late reply.
It was me that was not following correctly.
I did not catch the bit about the implicit default value insertion of MySQL, which led me to the wrong train of thought.

I'm with you on the point that alllowing fields to be empty that weren't before could lead to strange behaviour and this is the safer way to continue until someone can give it a try and fix all implicit assumptions of existing default values.

I'll merge this now as it blocls any new installation to succeed

M4LuZ:
Sadly we have to go for a code change, as the default config enforces both strict mode and `NO_ZERO_DATE`, raised #473 to fix that for the birthday field, (many) more to follow

PR #461, opened 2020-11-21
TravisCI: Exclude PSR1.Methods.CamelCapsMethodName.NotCamelCaps rule

andygrunwald:
We exclude the PSR1.Methods.CamelCapsMethodName.NotCamelCaps rule to make the build pass again.
The codebase is quite old and it might not lead to a huge benefit of
changing all the method's name to match the NotCamelCaps rule.


M4LuZ:
Sorry, late to the party.
Fine by me, there is more important stuff to worry about.
(would have voted though to name branch "make-master-great-again" ;))

PR #463, opened 2021-01-03
Fix inserts in translations table

t-h-e:
see code change in modules/install/Classes/Install.php
Length of $org and other translations need to be checked for size
 restrictions (255 chars), as they may be longer than $org

also fixed UTF-8 encoding issues

fixes parts of #456


modules/tournament2/mod_settings/translation.xml
```diff
@@ -2294,7 +2294,7 @@
 			</entry>
 			<entry>
 				<id>dead631430c69fb1151f9d0330e80516</id>
-				<org>Neues Team (Spieler) anmeldenHTML_NEWLINE(Nur in Anmeldephase m�glich)</org>
+				<org>Neues Team (Spieler) anmeldenHTML_NEWLINE(Nur in Anmeldephase müglich)</org>
```
andygrunwald, line 2297:
```suggestion
				<org>Neues Team (Spieler) anmeldenHTML_NEWLINE(Nur in Anmeldephase möglich)</org>

```

andygrunwald (changes requested):
@t-h-e Thanks a lot.
In general, it looks pretty good. I left a small comment, before merging.

M4LuZ:
Requested change was done, thus OK and fine to merge

M4LuZ (approved):

PR #464, opened 2021-01-03
Fix TranslationTests in php 7.4

t-h-e:
travis failed in other PRs


M4LuZ (approved):
Ahh yes, another one of those, thanks for the fix

PR #465, opened 2021-01-08
Check if function get_magic_quotes_gpc to avoid PHP Fatal error in PHP > 7.4

andygrunwald:
See ticket #450

### DIFF
--- a/index.php
+++ b/index.php
@@ -229,8 +229,8 @@ $__POST = $_POST;
 // Emulate MQ, if disabled
 // TODO Remove this get_magic_quotes_gpc function check, once this project had 7.4 as a minimum requirement.
 // See
-//	- Setting: https://www.php.net/manual/en/info.configuration.php#ini.magic-quotes-gpc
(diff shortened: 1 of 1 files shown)


Dwarfex:
the function that is checked for existance here was removed with php 5.4.0 - i think you could remove the check as a whole as 7.0 'is' and 7.3 'should' be minum required

andygrunwald:
@Dwarfex Good point.
I added a TODO to remove it. I did not remove the check yet, because
* LANSuite min. the requirement is PHP 7.0
* The setting `magic_quotes_gpc` was deprecated in PHP 7.3 and removed in PHP 7.4
* The function `get_magic_quotes_gpc` was deprecated in PHP 7.4 and removed in PHP 8

By this, we aim to keep the same functionality without breaking anything until we raise the min. PHP requirements.

Additional note: magic_quotes_gpc is a setting that we should not use anyway, same for the addslashes behavior in this context. However, right now we miss the time to fix this for the application.

What do you think on this?

M4LuZ (approved):
Tested against local setup with deactivated party class - OK

PR #467, opened 2021-01-08
Prepare for PHP7.3, 

Dwarfex:
Hi, 
yes i know this PR is a bit long. I put each type of change in a separate commit to make it easier viewable.

TLDR: I fixed some potential bugs, replaced (soon to bee) deprecated functions 

After i run phpstan, phan and PHPStorm Inspections, there was a huge list and i started to fix some(in my perspective) major ones.

- Security: allow only de-serialization to ProductList::class
- fix bug where "get_party_dropdown_form" would always show old (disregarding $show_old parameter)
- fix possible bug where LanSuite\Module\CashMgr\Account::class would disregard $party_id
- removed identical, duplicated array entries
- added used, but not mentioned extensions to composer file
- remove setting magic quotes to "0" as its always off (doesn't exist anymore)
- replace deprecated 'while = each()' in favor for 'foreach'
- replace eregi() with preg_match() (case insensitive)
- replace function calls with inbuilt constants
- replace alias function calls with originals
- performance (x2): replace 'array_push($array, $value)' with '$array[] = $value';
- replace 'for' loops with object counting in favor for 'foreach'
- clarified parentheses(added brackets for "how php does interpret"): needs validation if this is actual wanted behavior
- joined switch statements that do exactly the same for better overview

Let me know if there are any changes - i'm happy to rework if needed - so this can be merged!
P.S.: Yes - i changed files in /ext_scripts on purpose as i don't know what is really used there. (thats some work for the future i'd guess ;) ) 

Greetings,
Dwarfex

PR #468, opened 2021-01-09
Add GitHub Actions workflow to replace TravisCI

andygrunwald:
This PR replaces TravisCI with GitHub Actions.

### Why?
- TravisCI itself is moving more and more away from open source support after an acquisition, see https://news.ycombinator.com/item?id=25338983
- GitHub Actions is accessible for everyone and built into our source code platform (easy to use)
- GitHub Actions scheduling is often faster than TravisCI

### What is checked

- PHP Unit for PHP 7.2, 7.3, 7.4 and 8
- PHP Code Sniffer for all files

### What is still TODO after merging

- [ ] Configure the PHP Unit jobs as [required for merging](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/enabling-required-status-checks) (the CodeSniffer checks are failing at the moment, due to [missing "allow-failure" for a given job support](https://github.com/actions/runner/issues/2347))
- [ ] Remove TravisCI Webhook from the repository settings
- [ ] Deactivate the lansuite repository in the TravisCI Service


andygrunwald:
Hey @M4LuZ @Dwarfex,
would love to get your quick review on this.

Dwarfex:
Hi @andygrunwald 
my humble ideas to improve: 
- add PHP 8.0 to composer for UnitTesting ( "php": "^7.0 | ^ 8.0", )
- only inspect for code style in the "real" project folders - currently also /vendor and all other folders are checked
-  maybe let phpcbf run and commit the result? (so contributors are helped with the issue of code style and they only have to fix the hard erros)

Greetings,
Dwarfex

andygrunwald:
@Dwarfex that are great suggestions. Thanks for this. Let me go through those:

> - add PHP 8.0 to composer for UnitTesting ( "php": "^7.0 | ^ 8.0", )

This would be really good. Is there an option to only allow this for unit testing?
If we raise the PHP requirement, then we would allow lansuite on PHP 8 already. I have the feeling that the app is not ready for this version yet (we are preparing for proper PHP 7.3 support https://github.com/lansuite/lansuite/pull/467)

> - only inspect for code style in the "real" project folders - currently also /vendor and all other folders are checked

True. I applied an exclude of the vendor folder in https://github.com/lansuite/lansuite/pull/468/commits/8f26cc4a5e8132b5b88823c42038a014580e16f1.
Instant speedup. I will keep all other folders for now (since only files with the ending of PHP will be supported)

> - maybe let phpcbf run and commit the result? (so contributors are helped with the issue of code style and they only have to fix the hard errors)

Also a good idea. And in general, I am a fan of those automatic fixes (even inside CI).
However, for now (and mainly for simplicity), I would be in favor of making it easy to apply those fixes on the client and commit those. E.g., by a proper make command or something.

Dwarfex:
@andygrunwald 
>  Is there an option to only allow this for unit testing?
To my current knowledge there this requirement is global for the project and not per environment (as the other dev packages).
Nevertheless i think it would be best to remove all the (already) deprecated functions in PHP7.3, then the application would also run on PHP 8.0 
I could check this in an additonal PR/Issue.

> proper make command or something. (for phpcbf)
Maybe included your current phpcs/phpcbf commands as an easy to run composer command?
Like the phan and phpstan command.

andygrunwald:
> EDIT: I just checked - i removed the deprecated functions with the PR for 7.3

Thanks a lot. I am still reviewing it

> Maybe included your current phpcs/phpcbf commands as an easy to run composer command?

That is a good idea. Can you create a PR for this?

.github/workflows/continuous-integration.yaml
```diff
@@ -0,0 +1,77 @@
...
+        with:
+            path: ${{ steps.composer-cache.outputs.directory }}
+            key: ${{ matrix.php }}-composer-${{ hashFiles('**/composer.lock') }}
+            restore-keys: ${{ matrix.php }}-composer-
+
+      - name: Install dependencies
+        run: composer install
+
+      - name: Run tests
+        run: bin/phpcs --ignore=*/vendor/* --standard=PSR1,PSR2 --exclude=Generic.Files.LineLength,PSR1.Methods.CamelCapsMethodName --extensions=php -s .
```
M4LuZ, line 77: This appears to be a functional regression (and also causing the check to fail)
Instead of checking modified files, it appears that syntax check is applied to all files, which leads to longer execution time and also the problem that syntax has to be fixed everywhere for this to pass

M4LuZ: modified part to only call PHPCS on changed files

M4LuZ (approved):
Should be fine now, happy to have this merged

M4LuZ:
Merging this now to ensure rough coverage till @andygrunwald can have another look

PR #472, opened 2022-03-14
Fix copy and edit seat plan

t-h-e:
"orientation" and "u18" in seat_block are `enum` fields and not an `int`. Therefore, single quotes are required and the field should simply be treated as %string% in lansuite database queries.

"u18" is a checkbox and no value is sent on form submit if it is unchecked. Before "u18" was treated as `int` and PHP converted an empty string to 0 which old versions of MySQL accepted. Now MySQL is not as forgiving and "u18" has to be a string, but an empty string is not accepted, therefore special handling is needed for u18.

Similar to Import.php and edit.php

### DIFF
--- a/modules/seating/copy.php
+++ b/modules/seating/copy.php
@@ -24,8 +24,8 @@ switch ($_GET['step']) {
             `rows` = %int%,
             cols = %int%,
             `name` = %string%,
-            orientation = %int%,
-            u18 = %int%,
+            orientation = %string%,
+            u18 = %string%,
             remark = %string%,
             text_tl = %string%,
             text_tc = %string%,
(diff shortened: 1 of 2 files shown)


modules/seating/copy.php
```diff
@@ -44,7 +44,7 @@
             $row['cols'],
             $row['name'] .' (Kopie)',
             $row['orientation'],
-            $row['u18'],
+            empty($row['u18']) ? $row['u18'] : '0',
```
andygrunwald, line 47: Is this check correct?
It says If $row['u18'] is empty, use $row['u18'], otherwise 0.

Why should we use $row['u18'], if $row['u18'] is empty?

andygrunwald (changes requested):
Looks good, except of my one question

modules/seating/copy.php
t-h-e: Good catch. This is the wrong way around.
Fixed it.

andygrunwald (approved):

andygrunwald:
Thanks @t-h-e

PR #474, opened 2022-03-27
Change code to allow empty birthday field

### LINKED ISSUE #473: Implicit use of zero date for birthday causes issues with MySQL strict mode and no_zero_date
This is one of many related issues.
Currently the user.birthday field is assumed to be set to '0000-00-00' when no value is entered.
This is used for comparison in different parts of the code, but collides with the strict mode newer MySQL versions enforce by default

## Expected Behavior

Field should not contain a value if not set.
Code should check for empty values first and foremost, fallback would be to compare against zero dates

## Current Behavior

Comparisons are done by running `$row['birthday'] == '0000-00-00`

## Possible Solution

Change DB field be nullable, change inserts to write no value if none given.
Modify all checks for birthday field to compare for empty values and zero date

Try to install fresh on MySQL 5.7 ;)

## Your Environment

* Version used: Git master
* Operating System and version: Debian Bullseye, PHP 7.4, MySQL 5.7
* Enabled features:

M4LuZ:
Validated change on fresh PHP 7.4/MySQL 5.7 install.

Only incompatiblity found: Editing the user sets empty birthday field to <current date - 80 years>, but that is also behaviour as before. 

fixes #473

### DIFF
--- a/modules/install/wizard.php
+++ b/modules/install/wizard.php
@@ -35,7 +35,7 @@ switch ($_GET["step"]) {
             // If not found, insert
             } else {
                 $db->qry(
-                    "INSERT INTO %prefix%user SET username = 'ADMIN', firstname = 'ADMIN', name = 'ADMIN', email=%string%, password = %string%, type = '3', lastlogin = NOW(), comment = '', birthday = '0000-00-00', signature = ''",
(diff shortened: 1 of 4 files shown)


andygrunwald (approved):
LGTM. Feel free to merge it whenever you are ready

PR #481, opened 2022-04-25
415 rework translations

### LINKED ISSUE #415: Translation loading broken for generic/global translations

M4LuZ:
WIP - change of translation module to use caching, refactor function and fix issue with language selection during initial setup

fixes #415


M4LuZ:
Closing this, as functionality is provided with #650.
Caching part will be raised with new PR at some point

PR #483, opened 2022-05-06
Initial draft of headers

### LINKED ISSUE #179: Add HTTP security headers
## Expected Behavior

Over the last years a good amount of security enhancements have been included into the HTTP standard. Some of them are configured by the server sending defined HTTP header bits.
Thus we should have a look if we can implement these without breaking current functionality.
Examples would be:
* X-Frame-Options
* X-XSS-Protection
* X-Content-Type-Options
* Strict-Transport-Security 

More to be found for example at the [OWASP Secure Headers Project](https://www.owasp.org/index.php/OWASP_Secure_Headers_Project#tab=Headers)
## Current Behavior

Headers aren't set. Default Browser options are used

## Possible Solution

Set headers per default to recommended setting, fix compatibilty issues if possible, change header value otherwise

## Context

To get LS up to the current state-of-the-art in this regard

M4LuZ:
Includes settings for cookies

fixes #179

### DIFF
--- a/ext_scripts/captcha.php
+++ b/ext_scripts/captcha.php
@@ -9,7 +9,7 @@ function imageAuthCode($width, $height) {
   $auth_code = md5($auth_code);
   $auth_code = substr($auth_code, 0, 5);
   $auth_code = strtoupper($auth_code);
-  setcookie('image_auth_code', md5($auth_code), time()+60*60 , '/');
+  setcookie('image_auth_code', md5($auth_code), time()+60*60 , '/',"",$_SERVER['HTTPS'] = 'on' ? true : false, true);
   $image = imagecreate($width,$height) or die("Can't initialize GD image stream");
 
   $bg_color     = imagecolorallocate($image, 240, 248, 255);
--- a/inc/Classes/Auth.php
+++ b/inc/Classes/Auth.php
@@ -676,7 +676,9 @@ class Auth
             $this->cookiedata_pack(),
             time()+3600*24*$this->cookie_time,
             $this->cookie_path,
-            $this->cookie_domain
+            $this->cookie_domain,
+            ,$_SERVER['HTTPS'] = 'on' ? true : false, 
+            true
         );
     }
 
--- a/index.php
+++ b/index.php
@@ -95,6 +95,15 @@ if (isset($frmwrkmode)) {
 
 // Set HTTP-Headers
 header('Content-Type: text/html; charset=utf-8');
+header('X-Frame-Options: sameorigin');
+header('X-XSS-Protection: 1; mode=block');
+header('X-Content-Type-Options: nosniff');
+header('Referrer-Policy: no-referrer');
+header("'Content-Security-Policy: default-src 'self'; script-src 'self'; img-src 'self'; style-src 'self'; font-src 'self'; object-src 'none'; frame-src 'self'; worker-src 'self'; connect-src 'self';");
+// Enforce HSTS if browsing via HTTPS
+if ($_SERVER['HTTPS']=='on') {
+    header('Strict-Transport-Security: max-age=86400; includeSubdomains');
+}
 
 include_once("ext_scripts/mobile_device_detect.php");
 $framework->IsMobileBrowser = mobile_device_detect();


M4LuZ:
@lansuite/lansuite-developers : It would be good you give the proposal a try.
_Should_ be fine, but would appreciate your feedback

M4LuZ:
_(removing from milestone, as not critical and no feedback received so far)_

andygrunwald:
I am planning to review, check and test this one soon.

M4LuZ:
I think some additional work needs to be done on this, as current values prevent external inclusions that would be needed for gmaps, analytics and I think also discord.
But I guess that should be doable by providing the option to modify settings from standard values when needed (potentially by changing related code to use HttpFoundation?)

andygrunwald:
> I think some additional work needs to be done on this, as current values prevent external inclusions that would be needed for gmaps, analytics and I think also discord.

I am sure there is a right set of attributes to allow / whitelist external includes. Let me check a bit about this.

> But I guess that should be doable by providing the option to modify settings from standard values when needed (potentially by changing related code to use HttpFoundation?)

I may not understand how `HttpFoundation` helps here.

ext_scripts/captcha.php
```diff
@@ -9,7 +9,7 @@ function imageAuthCode($width, $height) {
   $auth_code = md5($auth_code);
   $auth_code = substr($auth_code, 0, 5);
   $auth_code = strtoupper($auth_code);
-  setcookie('image_auth_code', md5($auth_code), time()+60*60 , '/');
+  setcookie('image_auth_code', md5($auth_code), time()+60*60 , '/',"",$_SERVER['HTTPS'] == 'on' ? true : false, true);
```
andygrunwald, line 12: For the HTTPS detection, we should use `isSecure()` by HttpFoundation, see https://github.com/symfony/http-foundation/blob/439fdfdd344943254b1ef6278613e79040548045/Request.php#L1043

It respects also proxies and loadbalancers.
I know, this might be tricky in the captcha here, but maybe we should switch to another captcha lib in this phase.

inc/Classes/Auth.php
```diff
@@ -715,7 +715,7 @@ private function cookie_set()
         setcookie(
             $this->cookie_name,
             $this->cookiedata_pack(),
-            ['expires' => time()+3600*24*$this->cookie_time, 'path' => $this->cookie_path, 'domain' => $this->cookie_domain]
+            ['expires' => time()+3600*24*$this->cookie_time, 'path' => $this->cookie_path, 'domain' => $this->cookie_domain, 'secure' => $_SERVER['HTTPS'] == 'on' ? true : false, 'httponly' => true]
```
andygrunwald, line 718: See HTTPS detection comment above - https://github.com/symfony/http-foundation/blob/439fdfdd344943254b1ef6278613e79040548045/Request.php#L1043

index.php
```diff
@@ -114,6 +114,15 @@
 
 // Set HTTP-Headers
 header('Content-Type: text/html; charset=utf-8');
+header('X-Frame-Options: sameorigin');
+header('X-XSS-Protection: 1; mode=block');
+header('X-Content-Type-Options: nosniff');
+header('Referrer-Policy: no-referrer');
+header("'Content-Security-Policy: default-src 'self'; script-src 'self'; img-src 'self'; style-src 'self'; font-src 'self'; object-src 'none'; frame-src 'self'; worker-src 'self'; connect-src 'self';");
+// Enforce HSTS if browsing via HTTPS
+if ($_SERVER['HTTPS']=='on') {
```
andygrunwald, line 123: Same comment on HTTPS detection - https://github.com/symfony/http-foundation/blob/439fdfdd344943254b1ef6278613e79040548045/Request.php#L1043

index.php
```diff
@@ -114,6 +114,15 @@
 
 // Set HTTP-Headers
 header('Content-Type: text/html; charset=utf-8');
+header('X-Frame-Options: sameorigin');
+header('X-XSS-Protection: 1; mode=block');
```
andygrunwald, line 118: I am not sure if we should use this header. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

> Non-standard: This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

and

> Warning: Even though this feature can protect users of older web browsers that don't yet support [CSP](https://developer.mozilla.org/en-US/docs/Glossary/CSP), in some cases, XSS protection can create XSS vulnerabilities in otherwise safe websites. See the section below for more information.

index.php
```diff
@@ -114,6 +114,15 @@
...
+header('X-Content-Type-Options: nosniff');
```
andygrunwald, line 119: The idea of this header is great and I like the idea. However, according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options it says

> The X-Content-Type-Options response HTTP header is a marker used by the server to indicate that the [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types) advertised in the [Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) headers should be followed and not be changed.

I am not aware that we are sending proper mime types in every request. However, I see this as a pre-requisite to make use of this header (and keep this system working).

My suggestion: Commenting this header out for a later use once we got more confidence sending the right mime types

index.php
```diff
@@ -114,6 +114,15 @@
...
+header('Referrer-Policy: no-referrer');
```
andygrunwald, line 120: What is the purpose of sending no referrer at all?
Often, the referrer is pretty useful, especially when this website is sending bad requests to some other server.

I am more in favor of using the default policy (`[strict-origin-when-cross-origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#strict-origin-when-cross-origin)`) here.

index.php
```diff
@@ -114,6 +114,15 @@
...
+header("'Content-Security-Policy: default-src 'self'; script-src 'self'; img-src 'self'; style-src 'self'; font-src 'self'; object-src 'none'; frame-src 'self'; worker-src 'self'; connect-src 'self';");
```
andygrunwald, line 121: This might be tricky with the external libraries we include. We may need to specify domains on this one, see https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#example_3

I am not sure if this works.

What do you think about starting learn and only going with `default-src 'self'` at first?

andygrunwald (changes requested):
I had a deeper look and this might need a bit of more work. See my code comments.

index.php
M4LuZ: I have more confidence that the right mime types are being sent, as a mapping is done by the webserver anyways based on file extension unless specifically told not do do so by configuration or overridden.
LS does set MIME-type specifically for file downloads e.g. in the gallery (see modules/picgallery/download.php)

So I see no issue with setting that (and also was not able to find one in superficial tests), but fine to leave it out for now

index.php
andygrunwald:
```suggestion
```

index.php
andygrunwald:
```suggestion
// TODO: This header is still useful - Once we verified to send the correct MIME types, enable this header
// header('X-Content-Type-Options: nosniff');
```

index.php
andygrunwald:
```suggestion
header('Referrer-Policy: strict-origin');
```

andygrunwald:
> If you mean the reference to $request->isSecure() then that is a good find that I was not aware of, but not what I meant ;)

Yes, I mean using the `isSecure` method to check if the request is HTTPS or not.

> That is a yes if you meant the response object and the related options / functions to manipulate headers in the response
https://symfony.com/doc/current/components/http_foundation.html#response

Oh, we should use the lib to send the respective response headers. The challenge here: We don't use the response object at all right now.

> I'm in doubt if that is an easy implementation given the output buffer management done via LS and Smarty, so we may be better to use native functions or just a adequate wrapper for our management for now.

I am in line with you. My suggestion: Merging these headers once we clarified the details. Migrating to the response object at some later point in time. This might be a bigger change.

M4LuZ:
> I am in line with you. My suggestion: Merging these headers once we clarified the details. Migrating to the response object at some later point in time. This might be a bigger change.

Actually...
It appears that we can benefit from `httpFoundation` nevertheless and now by:

- creating (yet another) global object. This time of `ResponseHeaderBag` (this is an object that is used within the httpFoundation response to keep track of headers and cookies).
- Instantiate all headers that we want to set with standard values in `index.php`
- have modules modify these as/if needed (we may need to handle `Content-Security-Policy` differently, as that may need to have multiple addittions over multiple calls)
- before we start output we have just to call `->all()` to get all headers defined and then iterate and set them all (or potentially use a derived class where we put that in a function?)

andygrunwald:
@M4LuZ Your suggestion sounds reasonable. However, I am more in favor of

1. Getting this first implementation done
2. Opening a second PR and working on the `HttpFoundation`/`ResponseHeaderBag` integration

Why?

* Work was invested here already and is close to "mergable"
* Merging this provides already value for LanSuite
* Making this change later has no impact on the end user and is an internal change (ho additional effort needed by people who run LanSuite)
* Moving in small incremental steps is better and faster than "big bangs"
* Perfect is the enemy of good

M4LuZ:
I'm all in favor of getting a working solution out of the door, but current solution does not cover this still:

> [...] as current values prevent external inclusions that would be needed for gmaps, analytics and I think also discord.

So we either need to use default values for CSP that allow basically everything (which is not a good idea) oder at least add header overrides to related functions. Overwriting the referrer-option may also be needed for gmaps, as I think one of the recommendations is to restrict use of API keys to a site (which then needs to be checked via client referrer)

Also I see the level of code a bit differently:
* this as-is: draft
* having some type of header-management: usable
* Using some standard framework: good
* not having to deal with this at all: perfect ;)

M4LuZ:
Yes, I misunderstood, sorry.
Jut merging proven-to-be-ok headers is surely fine!
I continue to fail to see how getting the others in a manageable state is anything even close to "perfect", but happy to work on that in the separate tickets.

Will then include suggestions and remove headers covered by the spin-off tickets.

andygrunwald:
> I continue to fail to see how getting the others in a manageable state is anything even close to "perfect", but happy to work on that in the separate tickets.

With manageable you mean that modules can modify these headers?

M4LuZ:
We need at absolute minimum to have a modified CSP in place for:
* google maps
* google analytics
* ~discord~ (actually does not include external ressources)
So either these need to be added to the CSP in general or a dedicated CSP header call has to be defined and placed for these, otherwise inclusion of CSP will break functionality.

But that is no biggie, will put that on the to-do list with the other pending changes

M4LuZ:
Ahh, I see you did already most of the changes needed in the spin-off tickets.
Thus this can be closed and work continues in the separate tickets?

andygrunwald:
> Thus this can be closed and work continues in the separate tickets?

I suggest to first, merging the spin-off PRs and keep this open as a reminder to work on the CSP header.

PR #484, opened 2022-05-06
Remove module Teamspeak2, as it is certainly not in use anymore

M4LuZ:
LanSuite has a module for Teamspeak2, but it is definitely not in use by anyone anymore.
Either it will be replaced by the Teamspeak3 or Discord module or it will be simply gone.
There was a cross-dependency between the foodcenter module and TS2 (it used the refresh icons from there), which has been resolved also


M4LuZ:
@lansuite/lansuite-developers : Gents, could please one of you have a quick review of this?

BenniGotchi (approved):
Looks good!

PR #485, opened 2022-05-06
Implement new parser for proper BBcode tag evaluation

M4LuZ:
Replaces current parser for BBcode with a dedicated library.
It has been checked that this also does not open the door for XSS and solves the issue from #381

#fixes 381


M4LuZ:
@lansuite/lansuite-developers : Gents, could please one of you have a quick review of this?

M4LuZ:
@lansuite/lansuite-developers kind reminder

inc/Classes/Func.php
```diff
@@ -436,85 +436,92 @@ public function escape_sql($text)
     public function text2html($string, $mode = 0)
     {
         global $db;
+
```
andygrunwald, line 439: Can we have a quick documentation / inline comment about what `$mode` is and what possible values it can have?

Alternative: Replacing the magic numbers (0, 4, etc.) with constants that have a descriptive name.

andygrunwald (approved):
Just a small doc request to understand what $mode is.
Nothing blocking.

Thanks @M4LuZ!

inc/Classes/Func.php
M4LuZ: I'll try to remember what this was about and add a bit of documentation in that regard

Issue #183, opened 2018-04-02
Refactor database class to use PDO (+prepared statements) and to make it testable


andygrunwald:
## Expected Behavior

All database queries are using native prepared statements by the [PDO MySQL extension/functionality of PHP](https://secure.php.net/manual/en/ref.pdo-mysql.php).

## Current Behavior

LanSuite uses a "kind of" prepared statements, but with a self written method.
See https://github.com/lansuite/lansuite/blob/master/inc/classes/class_db_mysql.php#L179 for details.

## Possible Solution

Inject a PDO MySQL object into the database class via the constructor (dependency injections).
With this, we can make this class testable, because we would be able to mock the object.

Furthermore, use native prepared statements like described in the PHP docs.
For this, we would need to rewrite every query to use the new syntax.

## Context

To use the PHP standard way would sense.
Other software engineers will have an easier time to get familiar with the code.
Prepared statements are the standard of the database interaction and well known to be secure.

### More resources

* [The 2018 Guide to Building Secure PHP Software: Database Interaction](https://paragonie.com/blog/2017/12/2018-guide-building-secure-php-software#secure-php-databases)
* [Are PDO prepared statements sufficient to prevent SQL injection? @ StackOverflow](https://stackoverflow.com/questions/134099/are-pdo-prepared-statements-sufficient-to-prevent-sql-injection/12202218#12202218)

## Your Environment

* Version used: current master (rev. fd582f8394fbda39bee2de8e8a98ffc8c5e2e99c)

andygrunwald:
I found a related ticket in the old bugtracker. See http://lansuite.orgapage.de/index.php?mod=bugtracker&bugid=1171&ms_page=0

I had a small look and found zero Sleep-Queries in our Codebase.
I checked further and found a valid SQL Injection in the current LanSuite master and several older released version.
This explains the ticket in our old bugtracker.

Up to know i will not name how this injection can be triggered for security reasons (github don't have private issues). But enabling a proper database abstractions with prepared statement can avoid this.

aj-gh:
Is going for _PDO MySQL_ a strict requirement, and if so, why?
_mysqli_ would already be there in the _DB_ class and I guess going for _mysqli_ prepared statements would be a rather minor initial change compared to a full-blown switchover to PDO. One could even slowly migrate any existing statements, possibly by providing any additionally needed functions of _mysqli_ via _DB_.

andygrunwald:
[...]

PR #488, opened 2022-05-07
Initial draft to transform existing query functionality to prepared statements.

M4LuZ:
Introduces function qry2stmt that transform LS-style queries (with `%int%`,`%string%`,`%plain%`) to a MySQL prepared statement with bound variables.

There are still some issues as described in #183 so very much WIP with help appreciated

#refs 183


mrflobow:
I had a look into your transforming function and I think the issue is more in the way the DB class currently works.
You first call a query then the result it's stored in the current instance (maybe sometimes not), sometimes out of the function.

A better approach in my opinion

wrapper function e.g. 

fetch_array($query,...$params)
     -> calls a prepared statement function or simple query function private in the class
     -> returns result set or false

As a additional improvement use this as singleton

DBService::getInstance()->fetch_array("SELECT * FROM xyz WHERE id = ?",$id)
(Depending on PHP version either with ssi.. etc or PHP8 without)

I believe creating a seperate class and migrating exiting might from time to time might be an option.


I suggest some optimizing (if you want to stay with your approach)

````
$query_stmt = str_replace('%string%','?',$query);
        $query_stmt = str_replace('%int%','?',$query_stmt);
        $query_stmt = str_replace('%plain%','?',$query_stmt);

        //due to the usage of %plain% to drop in parts of the SQL statement it is required to iterate through query and deal with each entry separately
        $query_split = preg_split('((%string%|%int%|%plain%))',$query,-1,PREG_SPLIT_DELIM_CAPTURE);
        $query_param=''; //type definition for prepare statement

        foreach ($query_split as $query_element) {
            switch ($query_element) {
                case '%int%': //insert placeholder and write statement element with iterator
                    $query_param .= 'i';
                    break;
                case '%string%': //insert placeholder and write statement element with iterator
                    $query_param .= 's';
                    break;
                case '%plain%': //insert content as-is with iterator
                    $query_param .= 's';
                    break;
            }
        }

$stmt = $this->link_id->prepare($query_stmt);
$stmt->bind_param($query_param,...$args); 
$ok = $stmt->execute();
....


```

M4LuZ:
Hi @mrflobow ,

>  You first call a query then the result it's stored in the current instance (maybe sometimes not), sometimes out of the function.

What do you mean by that?
Also I have some trouble understanding your approach here:
* to me it appears that `fetch_array` would be on the same abstraction level (and apparently even has the same parameters) as `qry` currently.
* The singleton pattern would be just a refactoring without a functional impact (if we store the connection in an object instance globally available or as static class attribute globally available does not make a change in how it works, does it?)
* as per comment in #183 (which I guess refers to something here): The return value of mysqli::get_result() is either a `mysql_result` object or boolean `false` (see https://www.php.net/manual/en/mysqli-stmt.get-result.php)

andygrunwald:
FYI: I rebased this PR against the latest master.

andygrunwald:
I am all in for supporting Prepared Statements and even making this the new default.
I am not really sure if we should try to change the DB-Class with the existing functionality to support Prepared Statements.

From my current gut feeling we would benefit from a new Database class with first-level Prepared Statement support. Also to support unit testing and other functionality from the start. And then migrate query by query. This will take a while, this is correct.
But we would not run into several different (edge) cases like described here.

### PDO vs. mysqli

Another question that comes into my mind: Which Prepared Statement implementation should we use? PDO or mysqli?
LANSuite depends already on mysqli, which makes it easy to implement.
PDO would be a newly required extension (which is part of the standard distribution of nearly every PHP package for every Operating system).

What I like In particular about the PDO-Implementation: Named-Parameters (`:name`), see https://www.php.net/manual/en/pdo.prepare.php

This would make the query easier to read, because you don't have to count the number of `?`.

A good comparison is available here: https://websitebeaver.com/php-pdo-vs-mysqli

I would like to get your opinion on this, before we move on.

andygrunwald:
See https://github.com/lansuite/lansuite/pull/641 as an additional proposal.

M4LuZ:
Closing this as most of it it has been made obsolete with PRs #641 and #662.
It should now be very easy to create a wrapper from DB to Database that also includes mapping of the query calls.
But it makes sense to start there with a fresh codebase instead of continuing here

PR #491, opened 2022-08-17
Update DB.php

mohnewald:
Added missing $database to mysqli_connect.
Added $dbport to mysqli_connect. (must be set in inc/base/config.php)

 $dbport must be added in the installer, too.

### DIFF
--- a/inc/Classes/DB.php
+++ b/inc/Classes/DB.php
@@ -110,11 +110,13 @@ class DB
         $user = $config['database']['user'];
         $pass = $config['database']['passwd'];
         $database = $config['database']['database'];
+        $charset = $config['database']['dbport'];
         $charset = $config['database']['charset'];
+	    
 
         // Try to connect to the database
         // Suppress error output, because mysqli_connect throws a PHP Warning once it is not able to connect
-        $this->link_id = @mysqli_connect($server, $user, $pass);
+        $this->link_id = @mysqli_connect($server, $user, $pass, $database, $dbport);
 
         if (!$this->link_id) {
             if ($save) {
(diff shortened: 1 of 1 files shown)


inc/Classes/DB.php
```diff
@@ -110,11 +110,13 @@ public function connect($save = false)
         $user = $config['database']['user'];
         $pass = $config['database']['passwd'];
         $database = $config['database']['database'];
+        $charset = $config['database']['dbport'];
```
andygrunwald, line 113: Is there a default value if this is not set in the config? `3306` would be a good value.

M4LuZ: It would even be better if not `$charset` would be overwritten, but `$dbport` ;)

inc/Classes/DB.php
```diff
@@ -110,11 +110,13 @@ public function connect($save = false)
...
         $charset = $config['database']['charset'];
+	    
 
         // Try to connect to the database
         // Suppress error output, because mysqli_connect throws a PHP Warning once it is not able to connect
-        $this->link_id = @mysqli_connect($server, $user, $pass);
+        $this->link_id = @mysqli_connect($server, $user, $pass, $database, $dbport);
```
M4LuZ, line 119: Whilst the code directly below selects the DB as next action I would also be in favour to do so directly on connect.
Not specifying the DB would require the DB user to have `USAGE ON *.*` as privilege, if I'm not mistaken. 
Which _may_ be an issue in shared environments, but _should_ be not.

Just not sure if resulting code does handle DB-selection related errors OK.
Was that tested?

inc/Classes/DB.php
M4LuZ: resolved with new commit

andygrunwald (approved):

PR #507, opened 2022-12-29
Fixed broken caching mechanism due to framework changes

M4LuZ:
Framework implemented multiple function-breaking changes without noting them in the changelog, such as...
* Renaming of Classes
* Change of Paths
* Rework of functions and general usage

This fixes the part in `index.php`

### DIFF
--- a/docs/development-cache.md
+++ b/docs/development-cache.md
@@ -29,17 +29,18 @@ Code example:
 // Import global object
 global $cache;
 
-// Check for existence of item in cache
-if ($cache->has('module.entry.id') {
-   $CachedData = $cache->get('module.entry.id');
-   } else if {
-   // Not in cache, thus generate content
-   $CachedData = expensiveFunction();
-   // And store it in cache so that the next iteration uses cache data
-   $cache->write('module.entry.id', $CachedData, <$customTTL>);
-   }
-   //Run processing of data in $CachedData
-   ...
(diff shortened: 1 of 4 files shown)


mrflobow:
I'm currently working to get the docker variant up and running i stumble about the caching issue as it's a road blocker to get the system up and running.

May I suggest to use some more one function approach.

```
$config = $cache->get('config', function (ItemInterface $item) {

    $default = [];
    
    if (file_exists('inc/base/config.php')) {
        $default = parse_ini_file('inc/base/config.php', 1);
        return $default;
    // Default config. Will be used only until the wizard has created the config file
    } else {
      
        $default['lansuite']['default_design'] = 'simple';
        $default['lansuite']['chmod_dir'] = '777';
        $default['lansuite']['chmod_file'] = '666';
        $default['lansuite']['debugmode'] = '0';
 .....
        return $default;
    }
});

```

As per symfony documentation either it will load set config or will execute the anonymous function to get the code.
So you may avoid double checking if a variable is set.

M4LuZ:
@mrflobow: I agree that using an anonymous callback function would be more elegant, but even more so would it be to form "proper" cache contracts.
But looking at the readability of the code the simple  construct may still be the best ;)

Thanks for taking this over in your PR

Thus closing this here as modifications were merged via  PR #509

PR #520, opened 2023-01-12
Limit lasthiturl to 100 chars as we expect it to fit a varchar(100)

aj-gh:
Fixes SQL errors when visited URLs are longer than 100 chars. We could resize the database field however I don't really see the point of storing URLs with excessive length.. and it could still overflow at some point unless we make it really huge.


M4LuZ (approved):
Looks fine to me (besides the style issues blocking merge).
Had a quick fright that we may have an SQL injection on our hands, but apparently the variable is already cleaned up at https://github.com/lansuite/lansuite/blob/152309f4c37b087ecafaa93b67de5a9511d79165/index.php#L125

Issue #490, opened 2022-07-20
E-Mail communication does not support STARTTLS


Schmurki:
Hi! First of all I would like to thank you for the effort everyone put in this project!

## Expected Behavior

I have setup the configuration for the mail module in order to be able to send out mails to the participants for example when a user forgot his password (happens very often).

The mail module should be able to send out mails.

## Current Behavior

I only receive error messages by the SMTP server (I tried gmail and gmx).

```
Ran into problems sending Mail. Response: 530 5.7.0 Must issue a STARTTLS command first. r10-20020a5d494a000000b0021e40019ae7sm4247317wrs.48 - gsmtp
Ran into problems sending Mail. Response: 502 5.5.1 Unrecognized command. r10-20020a5d494a000000b0021e40019ae7sm4247317wrs.48 - gsmtp
Ran into problems sending Mail. Response: 502 5.5.1 Unrecognized command. r10-20020a5d494a000000b0021e40019ae7sm4247317wrs.48 - gsmtp
Ran into problems sending Mail. Response: 530 5.7.0 Must issue a STARTTLS command first. r10-20020a5d494a000000b0021e40019ae7sm4247317wrs.48 - gsmtp
Ran into problems sending Mail. Response: 530 5.7.0 Must issue a STARTTLS command first. r10-20020a5d494a000000b0021e40019ae7sm4247317wrs.48 - gsmtp
Ran into problems sending Mail. Response: 530 5.7.0 Must issue a STARTTLS command first. r10-20020a5d494a000000b0021e40019ae7sm4247317wrs.48 - gsmtp
Ran into problems sending Mail. Response: 502 5.5.1 Unrecognized command. r10-20020a5d494a000000b0021e40019ae7sm4247317wrs.48 - gsmtp 
```

The module should be able to send out mails by applying the provided configuration.

## Possible Solution

By looking into the PHP file that takes care of the function I observed that no TLS/STARTTLS is supported and from what I have read so far the module would need additional libs that provide these methods. I am not a PHP dev. so I have to read/try out a lot in order to maybe find a proper solution. Maybe someone more experienced is able to adjust this module in an hour :)

## Steps to Reproduce (for bugs)

1. Configure the mail module with the according settings
2. Enable that "internet" mails will be sent out by the system
3. Sent a mail as "internet" mail to a registered user

## Your Environment

* Version used: 
*   PHP Version   7.2.34

    MySQLi-Extension  mysqlnd 5.0.12-dev 

[...]

PR #524, opened 2023-01-16
Fix SMTP Mail send #490 

mrflobow:
This commit fixes https://github.com/lansuite/lansuite/issues/490.
For accomplishing the goal I added the libary phpmailer (as also suggested).
No big issues with adding it to composer. 
I believe with using the library lansuite mail module gets future proof and can be used in wider server configuration.

Tested following combinations:
* No Auth against local docker smtp container ->Success
* With Auth against MailTrap - >Success
* With TLS against Outlook. - >Success

For public SMTP you can't send as different user so I added an optional configuration to set the sender instead of the user itself. Example could be used with noreply@something.com .
If the configuration is left blank it will use the user E-Mail instead.

Summary:

* Refactored Mail Module
* Used widely use PHPMailer for sending mails instead of reimplenting from scratch.
* Added new settings to make SMTP configurable as needed.
* PHP Composer lock has been updated.


mrflobow:
In general the support for TLS has been established and the possibile to change the default port from port 25.
We can discuss about the from, that maybe can get from some external global settings that I'm not aware off. Otherwise it would be confusing not to have the setting there.

I have to say setting up a sendmail smtp relay is not trivial if you want to use your existing smtp on a linux server.
However the function was already there only slightly improved.

Are you questioning if SMTP should be supported at all?

M4LuZ:
Sorry, let me rephrase that:
This was not about the implementation details - these are fine and I'm happy to merge as-is, but about the necessity and implications of it.

As we have generally the following use-cases/usergroups:
* Somebody wanting to run this on managed webspace: mail is usually transmitted via preconfigured sendmail instance and we're fine, as we are now
* Somebody wanting to run this on their own box: normally same applies, we are also fine
* Somebody that runs an thrown-together setup locally with W/XAMPP (also there mercury acts as local sendmail and needs separate config) in a local network as intranet system

That leaves the following two use-cases where this would help:
* somebody wants to run LanSuite, but does not have the technical skills to do create, maintain and supervise the required environment
* Enterprise setups with large networks between Webserver and Mailserver, requiring encryption in between
* the odd case in a hosted environment where an remote mailserver AND encryption is required, or where an intranet instance should send mail to external recipients

And I worry that this enables especially the first group to do more harm than good.
But I could also worry for nothing and to support remote servers and TLS is a fair one.

mrflobow:
@M4LuZ  thanks for providing your elaborated thoughts.

I don't believe the operator would need to use mail/smtp at all for internal communication. So maybe it would be good to disable this at all. 
Only thing I see necessary is using mail / smtp to send password resets or newsletters. This would reduce the use case to a noreply at something adress + sendmail/smtp relay.

Would that lower your concerns?

My guess e.g. in AWS you would be happy to use SMTP directly and from my exp. it's a pain to setup sendmail/other centos magic :D to forward to SMTP Relay. So I'm always happy to have this in an application.  Just my 2cents on this :-)

andygrunwald:
Thanks @mrflobow, for your work on this.

I am a big fan of using established libraries for standard use cases like sending emails.
I understand that many people prefer using their self-hosted sendmail or similar service. However, a big fraction of people like to delegate this to a SaaS like one of the Cloud Hyperscalers or Mailgun.

My plan is to bring this PR to the latest and mergeable state.
I think it enables more people to use the module and all current use cases are also covered.

This might enable better usecases in the future. At least, we rely on properly maintained PHP packages for the heavy lifting.

andygrunwald:
@mrflobow @M4LuZ I fucked this branch up. Sorry.
My goal was to rebase it and make it mergable again. Instead I pushed way to many commits.

I opened a new PR and cherry-picked the original changes.
Look at https://github.com/lansuite/lansuite/pull/658

This will be the Pull Request to make this a success :)

PR #526, opened 2023-01-18
183 mysqli prepared statement dev

mrflobow:
* Removed store_result as conflicts with get_result
* Updated Auth Query, please check if it fits. I modified with best knowledge.
* Rebased with Master


M4LuZ:
> store_result as conflicts with get_result

I think the call to  `store_result` was added to fix issues with partial receival of data at time of call to `get_result` if I remember correctly.
But since this is under development anyways let's merge and see ;)

PR #569, opened 2023-05-14
Remove support for `get_magic_quotes_gpc`

andygrunwald:
The function and the PHP ini setting are deprecated since a long time and have been removed from PHP already.

### [Function: `get_magic_quotes_gpc`](https://www.php.net/manual/en/function.get-magic-quotes-gpc.php)

> This function has been DEPRECATED as of PHP 7.4.0, and REMOVED as of PHP 8.0.0. Relying on this function is highly discouraged.

### [Setting: `magic_quotes_gpc`](https://www.php.net/manual/en/info.configuration.php#ini.magic-quotes-gpc)

> This feature has been DEPRECATED as of PHP 5.3.0 and REMOVED as of PHP 5.4.0.

### DIFF
--- a/index.php
+++ b/index.php
@@ -133,29 +133,6 @@ $_SERVER['QUERY_STRING'] = $func->NoHTML($_SERVER['QUERY_STRING'], 1);
 // TODO investigate why this is needed
 $__POST = $_POST;
 
-// Emulate MQ, if disabled
-// TODO Remove this get_magic_quotes_gpc function check, once this project had 7.4 as a minimum requirement.
-// See
-// - Setting: https://www.php.net/manual/en/info.configuration.php#ini.magic-quotes-gpc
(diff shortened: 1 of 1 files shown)


M4LuZ (changes requested):
Unless I am mistaken only the check but not the other code can be removed.
As the logic is:
* If magic_quotes are present and enabled then all data in the superglobal arrays is already escaped - nothing to do
* if not then iterate though every single entry and escape control characters
Removing the whole block would strip LS of all sanitisation of user controlled data input, which certainly is not the desired state...

andygrunwald:
The whole `addslashes` and Magic quotes story is/was only needed, due to the possibility of running SQL Injections in LanSuite.

We are working towards a migration to Prepared Statements in
- https://github.com/lansuite/lansuite/pull/641
- https://github.com/lansuite/lansuite/pull/488

Once this is done, I propose this change again.
Until then, I will close it.

PR #570, opened 2023-05-14
Upgrade of development environment to PHP 8.1 + MySQL 8

andygrunwald:
Updates the basic development env to the latest official supported versions:

* PHP: 7.2.34 -> 8.1.18
* Composer: 1.8.0 -> 2.5.5
* xDebug: 2.9.0 -> 3.2.1
* phpunit/phpunit: ^7.1 -> ^10.1"
* nginx: Pinned at nginx:1.23.4-alpine
* MySQL: 5.6 -> 8.0.33

The application is not fully usable/runnable in this configuration, but it starts.

### A discussion about version end of life

< PHP 8.1 is EOL. MySQL 5.7 is still "okayish" and is getting end-of-life on October 21, 2023.
MySQL 8 was released 5 years ago (April 19, 2018).

The application lansuite is not 100% running/workable but boots up.
Mainly because PHP 8.1 is getting way stricter than PHP 7.x. Especially when it comes to access to undefined variables and array keys that are not existing.

### Draft

This Pull request is primarily a draft to get an impression about the future of LanSuite.
I feel that a large chunk of the codebase has to be refactored to match and let it run with the current software versions.
I want to gather some feedback/thoughts on how others see this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,7 @@
-FROM php:7.2.34-fpm
+# Get composer
+FROM composer:2.5.5 as composer
+
+FROM php:8.1.18-fpm-bullseye
 
 COPY . /code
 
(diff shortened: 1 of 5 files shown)


M4LuZ:
Not sure what to review here yet?
I can tell from experience that MySQL 8.0 is problematic (and I will add a bit to that in the discussion ticket), but why directly set PHP dependency to 8.1 if it will cause us that many issues? For nonfailing checks, sure, but as dependency now? 
PHP 8.0 does appear to be unproblematic _so far_

andygrunwald:
@M4LuZ Fair point.
I iterated over this PR a bit and opened a new one - See https://github.com/lansuite/lansuite/pull/575

Lets move the discussion to the new PR.

Issue #547, opened 2023-03-12
DB update fails due to unquoted keywords


M4LuZ:
## Current Behavior

DB update fails as lansuite uses reserved MySQL keywords as column names in the database.
The DB updater does not escape the names and thus causes sql errors for these.

## Possible Solution

add  " ` " to the related function to escape all names by default

## Steps to Reproduce (for bugs)

hit db manager, wait, check log

## Context

found while investigating other issues

## Your Environment

* Version used:
* Operating System and version:
* Enabled features:

M4LuZ:
Further investigation shows that this is just the tip of the iceberg.
Issues found so far:
* NULL / NOT NULL are not set as by definition, all fields are by defailt NOT NULL, but in many cases without default values. DB update does also not correctly detect and apply changes for these
* Masterform does not eval checkboxes correctly. An unselected checkbox does not have value '0', causing issues where DB fields are set to ENUM '0','1' (e.g. in tournament)

PR #575, opened 2023-05-17
Upgrade of development environment to PHP 8.0.28 + MySQL 5.7.42

andygrunwald:
Updates the basic development env to the minimum versions of the latest officially supported versions:

* PHP: 7.2.34 -> 8.0.28
* Composer: 1.8.0 -> 2.5.5
* xDebug: 2.9.0 -> 3.2.1
* phpunit/phpunit: ^8.5
* smarty/smarty: ^4.0
* youthweb/bbcode-parser: ^1.6
* nginx: Pinned at nginx:1.23.4-alpine
* MySQL: 5.6 -> 5.7.42

### PHP + Package upgrades

PHP went from `7.2.34` -> `8.0.28`.
PHP 8.x was active supported until **26 Nov 2022** (5 months ago).
It will receive security support until **26 Nov 2023.**
The end is near and approaching fast, but it is the least disruptive option for now.

The packages

* phpunit/phpunit: ^8.5
* smarty/smarty: ^4.0
* youthweb/bbcode-parser: ^1.6

required an upgrade to be compatible with PHP v8.0.
We did not upgrade to the latest version of these packages.

Source: https://www.php.net/supported-versions.php

### MySQL

MySQL went from `5.6` -> `5.7.42`
MySQL 5.7 will go end of life on  **October 21, 2023**
The end is near and approaching fast, but it is the least disruptive option for now.

The biggest issue (for me, at least):
MySQL 5.7 doesn't offer an official ARM docker image. Development on a modern ARM machine (aka mac M1) is only possible with modifications.

Sources:
* https://www.mysql.com/de/support/eol-notice.html
* https://endoflife.software/applications/databases/mysql
* https://endoflife.date/mysql

### Additional context

This PR is a successor of an early attempt: https://github.com/lansuite/lansuite/pull/570

### DIFF
--- a/inc/Functions/MyErrorHandler.php
+++ b/inc/Functions/MyErrorHandler.php
@@ -99,7 +99,7 @@ function MyErrorHandler($errno, $errstr, $errfile, $errline)
                 type = 3,
                 description = %string%,
                 sort_tag = "PHP-Fehler"',
-            (int) $auth['userid'],
+            (int) ($auth['userid'] ?? 0),
             $err
         );
     }
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -8,12 +8,12 @@ on:
 jobs:
   unit-tests:
     name: Unit Tests
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.4']
+        php-versions: ['8.0']
 #        include:
 #        - php-versions: 8.0
 #          experimental: true
@@ -49,7 +49,7 @@ jobs:
 
   style-tests:
     name: Code Style Tests
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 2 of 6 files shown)


inc/Functions/MyErrorHandler.php
```diff
@@ -99,7 +99,7 @@ function MyErrorHandler($errno, $errstr, $errfile, $errline)
                 type = 3,
                 description = %string%,
                 sort_tag = "PHP-Fehler"',
-            (int) $auth['userid'],
+            (int) ($auth['userid'] ?? 0),
```
M4LuZ, line 102: Not sure this was supposed to be in here, but fine anyways

M4LuZ (approved):

inc/Functions/MyErrorHandler.php
andygrunwald: This is, because otherwise it throws an error that `userid` is not existing. It assumes an authenticated user already.

M4LuZ: That is a generic underlying issue that DB field definition is either not correct or not correctly applied.

Both can be the case, as there are many locations where the DB definition just assumes that the DB fills in a default value, but sadly also the db update functionality does function correctly and sets NOT NULL on all columns (see https://github.com/lansuite/lansuite/issues/547#issuecomment-1465239882),

PR #583, opened 2023-05-23
Rector: Raise Level set to UP_TO_PHP_56 and ignore PowToExpRector

andygrunwald:
## Context

We raise the Level set of Rector to `UP_TO_PHP_56`. See https://github.com/rectorphp/rector/blob/main/config/set/level/up-to-php56.php

## Undefined variables

This PR will fix a lot of Warnings in PHP 8.
PHP 8 throws a Warning once a variable is used but not defined.

## Dependency

This PR depends on https://github.com/lansuite/lansuite/pull/582
#582 needs to be merged first.

The commit of this PR is https://github.com/lansuite/lansuite/commit/9ee877902aba4834a7a56a4a9236f11b7a67c8b9

### DIFF
--- a/modules/noc/class_noc.php
+++ b/modules/noc/class_noc.php
@@ -24,6 +24,7 @@ class noc
 {
     public function MACtoIP($SearchMAC, $IP, $ReadCommunity)
     {
+        $IP = null;
         $walk = snmprealwalk($IP, $ReadCommunity, ".1.3.6.1.2.1.3.1.1.2");
 
         if ($walk) {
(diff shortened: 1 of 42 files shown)


M4LuZ (approved):

modules/noc/class_noc.php
```diff
@@ -24,6 +24,7 @@ class noc
 {
     public function MACtoIP($SearchMAC, $IP, $ReadCommunity)
     {
+        $IP = null;
```
M4LuZ, line 27: This appears to be not correct.
``$IP``  is passed as parameter and then set to null?

M4LuZ (changes requested):
Should be a minor fix

modules/noc/class_noc.php
andygrunwald: Fixed in https://github.com/lansuite/lansuite/pull/583/commits/8284f31913a4974eb88b6b3641bafdf0fc8a1741

M4LuZ (approved):

PR #584, opened 2023-05-23
Rector: Raise Level set to UP_TO_PHP_70

andygrunwald:
## Context

We raise the Level set of Rector to `UP_TO_PHP_70`. See https://github.com/rectorphp/rector/blob/main/config/set/level/up-to-php70.php

## Dependency

This PR depends on https://github.com/lansuite/lansuite/pull/583
#583 needs to be merged first.

The commit of this PR is https://github.com/lansuite/lansuite/commit/58529a168764d64fbbe01635d8658b61c08fe05b


modules/noc/class_noc.php
```diff
@@ -24,6 +24,7 @@ class noc
 {
     public function MACtoIP($SearchMAC, $IP, $ReadCommunity)
     {
+        $IP = null;
```
M4LuZ, line 27: It's back again ;)
Will remove it myself, together with the other instance new with this PR

M4LuZ (approved):

PR #586, opened 2023-05-23
Rector: Raise Level set to UP_TO_PHP_72

andygrunwald:
## Context

We raise the Level set of Rector to `UP_TO_PHP_72`. See https://github.com/rectorphp/rector/blob/main/config/set/level/up-to-php72.php

## `each()`

[`each()`](https://www.php.net/manual/en/function.each.php) is deprecated and gets removed in PHP 8.

## Dependency

This PR depends on https://github.com/lansuite/lansuite/pull/585
#585 needs to be merged first.

The commit of this PR is https://github.com/lansuite/lansuite/commit/b83ec8dabc9fcf833934cefd50755f2627d46aa6


modules/foodcenter/Classes/ProductList.php
```diff
@@ -108,7 +108,11 @@ public function add_product($id, $opt)
 
                 // If it is not the same product, get the last key
                 end($this->product);
-                $key_array = each($this->product);
+                $key_array[1] = current($this->product);
```
M4LuZ, line 111: This (and the other instance) looks strange.

I think these should rather be initialized at beginning of the method, shouldn't they?

M4LuZ (approved):
Approving this already, but leaving merge open.
Please look at the comment and merge then

modules/foodcenter/Classes/ProductList.php
andygrunwald: I agree that this looks strange, but the new logic reflects the same behavior like `each()`:

See the output in different PHP Versions: https://3v4l.org/p9Pof

See a comment in the PHP docs: https://www.php.net/manual/de/function.each.php#126076

andygrunwald:
@M4LuZ I will wait until you merge it to ensure you are OK with my comment above

modules/foodcenter/Classes/ProductList.php
M4LuZ: Still looks strange to me, but I trust the documentation ;)
Guess I also dated myself with this, did stop to follow language development actively at some point

PR #602, opened 2023-05-26
Authentication: Define authentication types

andygrunwald:
## What this PR does

LanSuite is constantly checking things like

```php
$auth['type'] >= 2
```

Reading the code is pretty hard to know what `2` stands for.
It should be easy to understand, especially when we are dealing with permissions.

This Pull Request defines basic constants that can be used after merging to replace those integer comparisons.
The levels are taken from
https://github.com/lansuite/lansuite/blob/0da49dc87d61b0807869c796adda1961d7028690/inc/Classes/Auth.php#L522-L551

## Future implementation

After merging, we can start using these constants and make the code more readable.


M4LuZ:
Hi @andygrunwald what is the rationale for the long static prefix ``LS_AUTH_TYPE_<something>``?
Is there a chance that we run into conflicts with other definitions? 
Asking due to the typing needed ;)

Would it make sense to either make these class constants or extend $auth with related access / check functions e.g. ``is_userlevel(LS_ADMIN)`` and/or `` has_priv('tournament')``?

andygrunwald:
My thoughts around the long constant names are:

* `LS_` prefix to make it clear and explicit that this constant is a LANSuite constant
* `LS_` prefix to avoid any naming conflicts with 3rd party systems or the PHP language itself
* `AUTH` to make it explicit that this constant belongs to the Authentication/Authorization subsystem
* `TYPE` is another category inside the `AUTH` subsystem. See `LOGIN` as a different example - Mainly to have one constant for one purpose and not use `LS_AUTH_LOGIN_LOGGED_IN` and `LS_AUTH_TYPE_USER`for the same purpose because they both represent the value `1`

Happy to talk about this or to be challenged.
E.g. a name like `LS_ADMIN` can be misleading. This can mean the level of the Admin permission or an ADMIN account itself?
Maybe `LS_LEVEL_ADMIN` is a bit more explicit? But will there be also a `LS_TOURNAMENT_LEVEL_ADMIN`?

> Asking due to the typing needed ;)

From my experience + perspective, an IDE provides you with automatic code completion. Hence, I did not consider the length an issue.

> Would it make sense to either make these class constants or extend $auth with related access / check functions e.g. is_userlevel(LS_ADMIN) and/or  has_priv('tournament')?

I think adding such helper functions is a very good idea and we should do it.
I see this idea as an extension, because having named constant to avoid "magic numbers" (e.g. what does 3 mean?) is a requirement for proper understanding the codebase.

M4LuZ:
OK, fine for me then

M4LuZ (approved):

PR #606, opened 2023-05-28
Remove dependency to apcu extension

andygrunwald:
## Context

LanSuite suggests to install the apcu PHP extension ([APC User Cache](https://www.php.net/manual/de/book.apcu.php#book.apcu)).

This PR is removing the dependency on apcu for the following reasons:

1. We don't use a single `apcu_` function to make use on the user cache
2. It adds complexity to the server setup for the people who want to run LanSuite
3. We don't have any performance issue right now, which would be a compelling reason to add `apcu_` functions
4. Only the configuration cache has used the apcu cache - However, in my setups, I never used apcu and never observed that the system is slow - especially in times of SSDs, Ramdisks, etc.

I suggest to make the setup and maintenance of LanSuite easier by removing it.

Furthermore, this is not a breaking change.

## Dependencies

This PR depends on https://github.com/lansuite/lansuite/pull/604.
https://github.com/lansuite/lansuite/pull/604 need to be merged before.

The single commit in this PR is https://github.com/lansuite/lansuite/commit/30f46ed8fce575daf1364b6a314beb85883abbdf

### DIFF
--- a/docs/development-cache.md
+++ b/docs/development-cache.md
@@ -22,7 +22,7 @@ Also there is no requirement that the value is always correct, as it is just use
 ### Code
 
 There is object named `$cache` available in the global scope.
-This implements [PSR-16](https://www.php-fig.org/psr/psr-16/) and either works via APCu (if module enabled) or with files in the temporary directory.
+This implements [PSR-16](https://www.php-fig.org/psr/psr-16/) and works with files in the temporary directory.
 Code example:
 
 ```
--- a/inc/Classes/Auth.php
+++ b/inc/Classes/Auth.php
@@ -28,10 +28,8 @@ class Auth
 
     /**
      * Cookie data
-     *
-     * @var array
      */
-    private $cookie_data = [];
+    private array $cookie_data = [];
 
     /**
      * Cookie name
--- a/inc/Classes/Auth.php
+++ b/inc/Classes/Auth.php
@@ -250,6 +250,13 @@ class Auth
                 );
             }
 
+            if (!$user) {
+                $user = [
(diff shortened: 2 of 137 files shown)


M4LuZ:
Sorry, but not seeing the argument for it.

> We don't use a single apcu_ function to make use on the user cache

We do. The APCu instance of the cache is used automatically if module is present.
Thus everything that uses the cache adaptor then also uses APCu as cache backend.
That was the charme of using the symphony implementation for that.

>    It adds complexity to the server setup for the people who want to run LanSuite

It does not, as it is not a hard dependency and not defined as requirement in composer. 
If you don't have the module, then it works with a file cache without any change needed.

>    We don't have any performance issue right now, which would be a compelling reason to add apcu_ functions

We don't have stellar performance and it is just a piece of the puzzle at the moment.
When testing LS with a reasonable amount of requests the response times were...not good at all.
See https://github.com/lansuite/lansuite/pull/436#issuecomment-470061536
Did some profiling and most of the time are spend in ``NoHTML`` and ``t()`` if I remember correctly.
There are pending changes to these that should have them benefit from a cache.

>  Only the configuration cache has used the apcu cache - However, in my setups, I never used apcu and never observed that the system is slow - especially in times of SSDs, Ramdisks, etc.

I agree with you that (APCu) caching does not provide a large benefit at this point in time with the current level of implementation. but there are reasons it makes sense to use it and I saw at least a few hosting providers that run with it ootb, so it is not that uncommon to be available.

andygrunwald:
Thats fair. Your argumentation beats mine :)
I will close this one.

PR #634, opened 2023-05-31
Workaround for incompatibilities with SQL Mode `ONLY_FULL_GROUP_BY` and `STRICT_TRANS_TABLES`

andygrunwald:
## Context

This Pull Request is based on the work from @M4LuZ inside https://github.com/lansuite/lansuite/pull/476

What it does:
* Check the SQL Mode of the database - look for `ONLY_FULL_GROUP_BY` and `STRICT_TRANS_TABLES`
* Use `sqlmode` from the database config during database connect to set one for the current database session
* During installation: If an incompatible SQL mode was found, strip them out and add the compatible one to the configuration and write it to disk

### Installation

Screenshot 2023-05-31 at 22 34 42

### Admin-Seite > System-Test

Screenshot 2023-05-31 at 23 03 26


inc/Classes/DB.php
```diff
@@ -377,4 +386,13 @@ public function DisplayErrors()
             $this->errors = '';
         }
     }
+
+    /**
+     * Sets the SQL Mode for this database session.
+     */
+    public function setSqlMode(string $sqlmode)
+    {
+        $sqlModeQuery = sprintf("SET SESSION SQL_MODE='%s';", mysqli_real_escape_string($this->link_id, $sqlmode));
```
M4LuZ, line 395: As per prior comment: escaping is not really needed here, but it doesn't hurt, fine to leave as-is

M4LuZ (approved):

PR #641, opened 2023-06-08
Database/MySQL: Add new class with first class Prepared Statement support

andygrunwald:
## Context

This PR adds a new class `Database()` and deprecates the old class `DB()`.
The new class has first-class support for prepared statements. Primarily to support the security efforts and protect us against SQL Injections. Furthermore, is this considered best practice in software engineering.

This implementation is based on MySQLi, which is already required for LanSuite.

In this PR, there are also two example queries implemented in the News module to showcase a possible usage.

## Design choices

* The new class is not backward compatible / not a drop-in replacement - It is meant to be as a step-by-step migration to replace the old `DB()` class over time
* The new class does not support things like `print_error()` or debug capabilities (like adding the records to `log` table - Primarily because 
* The new class does not implement ALL functions of the previous one yet. `mysqli_fetch_array`is missing. The reason: We should prefer working with `assoc()` functions due to a more readable code by accessing the fields by name rather than number (0, 1, ...) - Assoc function is implemented in the new Database class
* The new class does not implement ALL functions of the previous one yet. `mysqli_fetch_field_direct` is missing. The reason: This will be implemented once the function is needed (during migration)
* The new class does not escape any values like the previous one (e.g. via `$func->NoHTML`). The reason: The class itself does not know in which context the value is used. Depending on the context (HTML, CSS, JS, ...) the value need a different type of escaping
* The new class has no external (global) dependencies like `$auth`, `$func`, ... - This makes it easy to re-use and test
* The new class throws PHP Exceptions rather than returning an error via string - The caller (code that uses the database class) should handle the error

## TODO

* Writing unit tests

## Open Questions

1. Debug/Timing query measurement: Do we need it? If yes, I can take this and implement this in an additional PR (to not blow this one up)
2. Error reporting in the `log` table: Do we need it? If yes, I can take this and implement this in an additional PR (to not blow this one up)

## MySQLi vs. PDO

This implementation uses MySQLi.
PDO (https://www.php.net/manual/en/pdo.prepare.php) supports named parameters.
It is brought up here: https://github.com/lansuite/lansuite/pull/488#issuecomment-1570177788
Especially in long queries, naming can help a lot.

Queries with named parameters can look like
```php
$sql = 'SELECT name, colour, calories
    FROM fruit
    WHERE calories < :calories AND colour = :colour';
```

vs (now):

```php
$sql = 'SELECT name, colour, calories
    FROM fruit
    WHERE calories < ? AND colour = ?';
```

I am also happy to implement this as PDO / with named parameters.
Let me know what you think. (/cc @M4LuZ)

## Possible next steps

* Getting a first review on this Pull Request - If there is a positive feedback, I can move on and implement the unit tests

### DIFF
--- a/inc/Classes/DB.php
+++ b/inc/Classes/DB.php
@@ -2,6 +2,14 @@
 
 namespace LanSuite;
 
+/**
+ * Old Database implementation.
+ *
+ * Please use the new database implementation Database()
+ * See Database.php
+ *
+ * @deprecated
+ */
 class DB
 {
     private \mysqli|bool|null $link_id = null;
--- a/inc/Classes/Database.php
+++ b/inc/Classes/Database.php
@@ -134,30 +134,18 @@ class Database
      * 
      * $query => The SQL statement.
      * 
-     * $parameterTypes => Parameter types:
-     *      i	corresponding variable has type int
-     *      d	corresponding variable has type float
-     *      s	corresponding variable has type string
-     *      b	corresponding variable is a blob and will be sent in packets
-     *      See https://www.php.net/manual/en/mysqli-stmt.bind-param
-     * 
      * $parameterValues => The values for the query.
-     *      Attention: Need to be in the same order as $parameterTypes.
      * 
      * TODO: Implement debug time measurement (with query_start and query_stop)
      */
-    public function query(string $query, string $parameterTypes = '', array $parameterValues = []): \mysqli_stmt
+    public function query(string $query, array $parameterValues = []): \mysqli_stmt
     {
         // Replace table prefix
         $query = str_replace('%prefix%', $this->tablePrefix, $query);
 
         $statement = $this->database->prepare($query);
-        if (count($parameterValues) > 0) {
-            $statement->bind_param($parameterTypes, ...$parameterValues);
-        }
-
         // Will throw an exception in the case of failure.
-        $statement->execute();
+        $statement->execute($parameterValues);
 
         return $statement;
     }
@@ -168,9 +156,9 @@ class Database
      * 
      * In case of an empty result, an empty array is returned.
      */
-    public function queryWithFullResult(string $query, string $parameterTypes = '', array $parameterValues = []): array
+    public function queryWithFullResult(string $query, array $parameterValues = []): array
     {
-        $statement = $this->query($query, $parameterTypes, $parameterValues);
+        $statement = $this->query($query, $parameterValues);
         $queryResult = $this->getStatementResult($statement);
 
         if (!$queryResult) {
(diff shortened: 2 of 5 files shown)


andygrunwald:
@M4LuZ Any possibility for a review in the near future?

M4LuZ:
> @M4LuZ Any possibility for a review in the near future?

I did have an initial look, but not done yet.
What I can place in advance is the question why parameter type binding is being used.
I had the same question (use it or not?) in a different team and as far as I remember it provides no benefit in regards to security

andygrunwald:
@M4LuZ How do you use prepared statements with MySQLi without adding the types in PHP 8.0?

I know the PDO Api doesn't need the types, but as asked above, this would require a new extension.

When we are moving towards PHP 8.1, we can drop the types because the https://www.php.net/manual/de/mysqli-stmt.execute.php supports parameters.
In PHP 8.0 this is not supported.

M4LuZ:
> @M4LuZ How do you use prepared statements with MySQLi without adding the types in PHP 8.0?

Errrm. 
Good question, would need to (re)search for that again.
I think it was done by treating everything as string

M4LuZ:
It appears that I had this correctly in mind. 
It is possible to just treat all parameters as string.
If defining data types is a requirements that will be gone again with 8.1, should we then define the new interface with them?
Or go for 'string' for now and then replace once we up requirements to 8.1?

> You can default to 's' for most parameter arguments in most cases.  The value will then be automatically cast to string on the back-end before being passed to the Mysql engine.  Mysql will then perform its own conversions with values it receives from PHP on execute.  This allows you to bind not only to larger numbers without concern for precision, but also to objects as long as that object has a '__toString' method.

> This auto-string casting behavior greatly improves things like datetime handling.  For example: if you extended DateTime class to add a __toString method which outputs the datetime format expected by Mysql, you can just bind to that DateTime_Extended object using type 's'.  [...]

https://www.php.net/manual/en/mysqli-stmt.bind-param.php#102917

andygrunwald:
@M4LuZ ALternative idea:
We raise the minimum version of LanSuite to PHP 8.1 already.
Why?
- PHP 8.0 active support was dropped on 26 Nov 2022 (7 months ago)
- PHP 8.0 only receives security updates until 26 Nov 2023 (4 months from now)
- PHP 8.1 was released 25 Nov 2021 (1 year, 7 months ago)
- This way, we don't need to iterate on this approach, but can make it ready from the start

We could also argue to raise the LanSuite minimum version of PHP to 8.2, because PHP 8.1 active support ends on 25 Nov 2023 (in 4 months), but it think for prepared statements this is OK for now.

What do you think?

M4LuZ:
>What do you think?

that is fine by me....if we do not end in massive compatibility or dependency issues.
If it is a quick change, then we should definitely go for that. 
Otherwise I would say we modify this PR and then spend the time needed for 8.1

andygrunwald:
@M4LuZ PHP 8.1 upgrade is here: https://github.com/lansuite/lansuite/pull/662
Once this is merged, I will update this PR.

M4LuZ:
>  Once this is merged, I will update this PR.

Done ;)

andygrunwald:
@M4LuZ I updated this PR and removed the types for the prepared statements. See https://github.com/lansuite/lansuite/pull/641/commits/1663924016b437d37850c01f2885ac82d726b111

M4LuZ (approved):

PR #650, opened 2023-06-21
Fix #344: Switching to English during Installation breaks the Installation page

### LINKED ISSUE #344: Setting Language to English Breaks Website
Hi, so after succesfully installing, I opened up the website, and changed the language from German to English, this made the website stop functioning, and throwing a lot of PHP errors.

Here's a imgur gallery with the php errors.
https://imgur.com/a/mUMN4RD

## Expected Behavior

Changing language to English should change language to English

## Current Behavior

Selecting English breaks the website.

1. Start fresh server
2. On setup page change to english and hit apply.

## Your Environment

* Version used: Latest
* Operating System and version: Ubuntu Server 18.04 
* Docker Version: 18.06.1-CE
* Docker Compose Version: 1.22.0

andygrunwald:
## Context

This Pull Request fixes the language selection during Installation.
A lot of translations are not available in English, but this should be fixed at a later stage / in a separate pull request.

Furthermore, we have a bug if you switch LanSuite to english itself.
The configuration (database configuration included) is loaded AFTER the language is selected.
I made a TODO in the comment and this is not fixed in this Pull Request

## Tickets

Fixes #344
Relate to #415

## Previous work

https://github.com/lansuite/lansuite/pull/481 is somehow related to this PR, but is not complete.

## Dependencies

This PR depends on https://github.com/lansuite/lansuite/pull/649
Please merge https://github.com/lansuite/lansuite/pull/649 before.


inc/Classes/Translation.php
```diff
@@ -163,13 +167,49 @@ private function load_cache_byfile($module)
...
+     *
+     * Does not overwrite existing language cache entries.
+     */
+    private function setLangCacheEntry(string $module, string $id, string $text)
+    {
+        if (!array_key_exists($module, $this->lang_cache)) {
+            $this->lang_cache[$module] = [];
+        }
+
+        if (array_key_exists($id, $this->lang_cache[$module]) && this->lang_cache[$module][$data['id']] != '') {
```
andygrunwald, line 186:
```suggestion
        if (array_key_exists($id, $this->lang_cache[$module]) && this->lang_cache[$module][$id] != '') {
```

inc/Functions/Environment.php
M4LuZ, line 1: Hmm. Should the filename not be equal to function name to adhere to PSR-4 (sort-of, as just for classes)?

inc/Functions/T.php
M4LuZ, line 1: What do you think about the approach of PR #481 to move the translation functionality from separate function t() to the related class `Translation`?
Not related to acceptance of this, but I considered that a good idea.

index.php
```diff
@@ -266,6 +266,12 @@
         $cache->save($cfgCache);
     }
     $cfg = $cfgCache->get();
+
+    // TODO Reload language
```
M4LuZ, line 270: As far as I can see / remember from the other PR the new language is passed as either $_GET or $_POST parameter and thus can be used to trigger cache flush / reload, see https://github.com/lansuite/lansuite/pull/481/commits/f9441a7e48e6e240aef68ad358e085bb5d1b719f#diff-e346d84a075b7d32e92ee9bd65971289b56a8fc7d80c836d68ac8afb19b72374

Could that be what is needed here?

M4LuZ (approved):
No major issues with code, see three comments before merging

inc/Functions/Environment.php
andygrunwald: Good catch. File renamed in https://github.com/lansuite/lansuite/pull/650/commits/0eedd84d7cf3e6e76d7c74009aad8b040a3be9a4

inc/Functions/T.php
andygrunwald: I saw this and liked the idea.
My plan is to rework the translation class a bit and move the function in a separate PR later as well. Primarily to make it reviewable.

index.php
andygrunwald: The reload is needed because the database is initialized AFTER the language is retrieved from GET/POST. The function you linked also checked for $cfg['sys_language'], which contains the configured language from the user.

A simple `get_lang()` was not sufficient during testing. This will set the language but don't reload the lang_cache.
I will look into this deeper, when we rework the translation class. Thats why I left https://github.com/lansuite/lansuite/issues/415 open

andygrunwald:
Merged, after approval. Comments taken into account.

index.php
M4LuZ: I meant this code block here within `get_lang()`
```
        if (isset($_POST['language']) && $_POST['language']) {
            $_SESSION['language'] = $_POST['language'];
            $this->lang_cache = array();
        } elseif (isset($_GET['language']) && $_GET['language']) {
            $_SESSION['language'] = $_GET['language'];
            $this->lang_cache = array();
```
The invalidates `lang_cache` which should then trigger it to reload it from source, which is then the DB

andygrunwald: Ah, I see.
I am not sure, if I prefer such a "side effect", because this function is called `get...` and not `clear...` or `refresh...`.
It won't be obvious that this function clears/resets the cache.

I would rather think about a method like "loadTranstrations()" or "refreshTranslations()" of something like this.

PR #654, opened 2023-06-23
Translation: Rework Translation class with typing, documentation and a smaller Interface

andygrunwald:
## Context

The translation class is key to the LanSuite system.
This Pull Request reworks the translation class a bit, **without changing any functionality**.

What we did:
* [Moved t() into Translation class](https://github.com/lansuite/lansuite/commit/e596458ce63229c201d9dbfa98d6943ad6201713)
* Provided a smaller public interface by making attributes and functions that are not in public use `private`
* Updated the documentation of the methods and properties
* Types the properties and methods (arguments and return types)
* Made the class a bit more configurable by things like [Translation: Replace 'de' with $defaultLanguage](https://github.com/lansuite/lansuite/commit/62b719809e2acaac01fab2efb95c436cbbaaa7a7)
* Made the code a bit more readable by removing obsolete `else` parts
* Fixed the one or other bug that referenced a wrong variable
* Added TODO comments for future work

The whole PR might be hard to review.
That's why I provided single commits for each isolated change.

## Next steps

The work in this class should continue.
We introduced a few TODO comments for future work, but my next steps are:

* Unit test the `translate()` method
* Check unit test coverage of the `Translation` class overall
* Remove the `global` usage in this class
* Refactor the `translate()` method with the goal of more simplified code

## Previous work

In https://github.com/lansuite/lansuite/pull/650 we have started the efforts on working on the translations. This pull request continues the efforts.

## Tickets

This is related to https://github.com/lansuite/lansuite/issues/415 (it is not fixing it, but kind of pre-work)

### DIFF
--- a/inc/Classes/Translation.php
+++ b/inc/Classes/Translation.php
@@ -266,15 +266,20 @@ class Translation
     }
 
     /**
-     * Read complete Module translation from database and write to back file.
+     * Read complete module translation from database and write to back file.
      * This will be only used for maintenance translations
      *
      * TODO Error handling for xml and file access
+     * TODO This function overwrites files in the source tree.
+     *      A LanSuite upgrade will overwrite the written file with a newer
+     *      version from the source code repository.
+     *      We should check if we want to support this functionality or modify it to
+     *      write a file outside of the source tree, to make it safe for code upgrades.
      *
      * @param string    $module     Module name
      * @return void
      */
-    public function xml_write_db_to_file($module)
+    public function xml_write_db_to_file(string $module): void
     {
         global $db;
 
@@ -283,7 +288,7 @@ class Translation
         $output = '<?xml version="1.0" encoding="UTF-8"?'.">\r\n\r\n";
         $header = $xml->write_tag('filetype', 'LanSuite', 2);
         $header .= $xml->write_tag('version', '2.0', 2);
-        $header .= $xml->write_tag('source', 'http://www.lansuite.de', 2);
+        $header .= $xml->write_tag('source', 'https://github.com/lansuite/lansuite', 2);
         $header .= $xml->write_tag('date', date('Y-m-d h:i'), 2);
         $header = $xml->write_master_tag("header", $header, 1);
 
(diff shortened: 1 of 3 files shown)


inc/Classes/Translation.php
```diff
@@ -4,25 +4,32 @@
...
+     * Global language.
+     */
+    private string $language = 'de';
+
+    /**
+     * Default language.
+     * Will be used in cases where an invalid language was selected.
      */
-    public $language = 'de';
+    private string $defaultLanguage = 'de';
```
M4LuZ, line 22: Hmm. If this is a value that is never changed...should we make this a constant?

andygrunwald: I would suggest the second option (making this configurable).
I created a ticket for this: https://github.com/lansuite/lansuite/issues/670

inc/Classes/Translation.php
```diff
@@ -32,28 +39,19 @@ class Translation
...
 
     /**
-     * Valid languages
+     * Valid languages.
      *
-     * @var array
+     * @var array<string>
      */
-    public $valid_lang = ['de', 'en', 'es', 'fr', 'nl', 'it'];
+    public array $valid_lang = ['de', 'en', 'es', 'fr', 'nl', 'it'];
```
M4LuZ, line 46: `$valid_lang = array_keys($this->lang_names);` ?

inc/Classes/Translation.php
```diff
@@ -300,79 +295,64 @@ public function xml_write_db_to_file($module)
...
-
-                if ($row['nl'] != '') {
-                    $entry .= $xml->write_tag('nl', $row['nl'], 4);
-                }
+            $entry = $xml->write_tag('id', $row['id'], 4);
+            $entry .= $xml->write_tag('org', $row['org'], 4);
 
-                if ($row['it'] != '') {
-                    $entry .= $xml->write_tag('it', $row['it'], 4);
+            $languagesToAdd = ['de', 'en', 'es', 'fr', 'nl', 'it'];
```
M4LuZ, line 316: Shouldn't this also be derived from the definition at top?

inc/Classes/Translation.php
```diff
@@ -300,79 +295,64 @@ public function xml_write_db_to_file($module)
...
-            }
-
-            if ($row['fr'] != '') {
-                $entry .= $xml->write_tag('fr', $row['fr'], 4);
-            }
-
-            if ($row['nl'] != '') {
-                $entry .= $xml->write_tag('nl', $row['nl'], 4);
-            }
+            $languagesToAdd = ['de', 'en', 'es', 'fr', 'nl', 'it'];
```
M4LuZ, line 347: same here

inc/Classes/Translation.php
andygrunwald: We could do this. However, I would prefer to get rid of `$this->lang_names` in the future at all.
Main reason: The language names are in german. If I aim to switch the language, the german language names doesn't help me much.
I would like to extract the language names into translation files in the future.

In this PR, I am to rework the class without big functionality changes, but rather make it type safe and limit the public interface.
In the next step, I would prefer to make such optimizations.

inc/Classes/Translation.php
andygrunwald: Fixed in https://github.com/lansuite/lansuite/pull/654/commits/d3bd13d050946bd78ce428444e760f3b7d0121fc

inc/Classes/Translation.php
andygrunwald: Fixed in https://github.com/lansuite/lansuite/pull/654/commits/d3bd13d050946bd78ce428444e760f3b7d0121fc

M4LuZ (approved):

PR #657, opened 2023-06-28
Module "tournament2": Fix several PHP Warnings and one PHP Fatal error

andygrunwald:
## Context

This PR is fixing several PHP Warnings in the `tournament2` module.

## Dependencies

This one depends on https://github.com/lansuite/lansuite/pull/653
https://github.com/lansuite/lansuite/pull/653 need to be merged before.

### DIFF
--- a/modules/tournament2/breaks.php
+++ b/modules/tournament2/breaks.php
@@ -3,7 +3,7 @@
 $dsp->NewContent(t('Turnier - Pausenverwaltung'), t('Trage Startzeitpunkt und dauer der Pause ein. Lansuite wird entsprechend spätere Rundenzeiten berechnen'));
 
 $stepParameter = $_GET['step'] ?? 0;
-switch ($_GET['step']) {
+switch ($stepParameter) {
     // Delete
     case 10:
         $md = new \LanSuite\MasterDelete();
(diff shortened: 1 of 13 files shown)


modules/tournament2/breaks.php
```diff
@@ -2,6 +2,7 @@
 
 $dsp->NewContent(t('Turnier - Pausenverwaltung'), t('Trage Startzeitpunkt und dauer der Pause ein. Lansuite wird entsprechend spätere Rundenzeiten berechnen'));
 
+$stepParameter = $_GET['step'] ?? 0;
 switch ($_GET['step']) {
```
M4LuZ, line 6: I think the intent to replace the next line with $stepParameter?

M4LuZ (approved):

modules/tournament2/breaks.php
andygrunwald, line 6:
```suggestion
switch ($stepParameter) {
```

andygrunwald: You are correct. Fixed in https://github.com/lansuite/lansuite/pull/657/commits/71930e5cc5033824b707ff6c7b09aa3c86df0d93

andygrunwald:
Merging through, because the approval was given.

PR #658, opened 2023-06-28
Introduce PHPMailer library and Fix SMTP Mail send

andygrunwald:
## Context

This Pull Request is the new version of https://github.com/lansuite/lansuite/pull/524.

While I wanted to make https://github.com/lansuite/lansuite/pull/524 mergable, I pushed to many changes to the PR. Hence, I created this new, clean one.

The kudos goes to @mrflobow.

Description from #524

> This commit fixes https://github.com/lansuite/lansuite/issues/490.
> For accomplishing the goal I added the libary phpmailer (as also suggested).
> No big issues with adding it to composer.
> I believe with using the library lansuite mail module gets future proof and can be used in wider server configuration.
> 
> Tested following combinations:
> 
> No Auth against local docker smtp container ->Success
> With Auth against MailTrap - >Success
> With TLS against Outlook. - >Success
> For public SMTP you can't send as different user so I added an optional configuration to set the sender instead of the user itself. Example could be used with [noreply@something.com](mailto:noreply@something.com) .
> If the configuration is left blank it will use the user E-Mail instead.
> 
> Summary:
> 
> Refactored Mail Module
> Used widely use PHPMailer for sending mails instead of reimplenting from scratch.
> Added new settings to make SMTP configurable as needed.
> PHP Composer lock has been updated.

## Tickets

Fixes https://github.com/lansuite/lansuite/issues/490
Related https://github.com/lansuite/lansuite/pull/524


andygrunwald:
I made a few additional fixes. Mostly PHP typing to make it consistent.

Next step:
* Setting up a proper dev environment for mail testing
* Testing the changes

modules/mail/mod_settings/config.xml
```diff
@@ -41,6 +41,18 @@
     		<type>string</type>
     		<default>localhost</default>
     		<description>Hostname des SMTP-Servers</description>
+    	</item>
+		<item>
+    		<name>mail_smtp_port</name>
+    		<type>int</type>
+    		<default>25</default>
```
M4LuZ, line 48: I would vote for going with Port {{465}} and TLS enabled as standard config.

M4LuZ:
Test cases I see:

- [ ] update case (switch from current codebase and config to this)
- [ ] unauthenticated localhost sending
- [ ] remote MTA with credentials
- [ ] TLS handling
- [ ] Error handling

modules/mail/mod_settings/config.xml
andygrunwald: This sounds like a good approach!
Let me look into this, once I got back into testing.

M4LuZ: resolved via 48a24ae4f6f4597f00d03925537a0e816341671b

PR #663, opened 2023-07-03
Scripts to build a reliable production release

andygrunwald:
## Context

A reliable way to build a production release based on Docker.

Why docker?
This way, we can ensure all folks involved build the same production release under the same version constraints.

## Documentation

A chapter was added to build the production release.

### DIFF
--- a/Dockerfile-Production-Release
+++ b/Dockerfile-Production-Release
@@ -1,7 +1,7 @@
 # Get composer
 FROM composer:2.5.8 as composer
 
-FROM php:8.0.29-fpm-bullseye
+FROM php:8.1.20-fpm-bullseye
 
 # Install libraries and PHP-Extensions that are not provided by the base image
 RUN apt-get update \
--- a/Dockerfile-Production-Release
+++ b/Dockerfile-Production-Release
@@ -1,0 +1,31 @@
+# Get composer
+FROM composer:2.5.8 as composer
+
+FROM php:8.0.29-fpm-bullseye
+
+# Install libraries and PHP-Extensions that are not provided by the base image
+RUN apt-get update \
+    && apt-get install -y \
+        libfreetype6-dev \
+        libjpeg62-turbo-dev \
+        libsnmp-dev \
+        snmp \
+        unzip \
+        libzip-dev \
+        git \
+    && docker-php-ext-install -j$(nproc) mysqli snmp \
+    && docker-php-ext-configure gd --enable-gd --prefix=/usr --with-jpeg --with-freetype \
+    && docker-php-ext-install -j$(nproc) gd \
+    # Cleanup
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-get clean
+
+# Composer setup starts here. The zip extension is required for that.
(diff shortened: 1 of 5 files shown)


Dockerfile-Production-Release
```diff
@@ -0,0 +1,31 @@
+# Get composer
+FROM composer:2.5.8 as composer
+
+FROM php:8.0.29-fpm-bullseye
```
M4LuZ, line 4: needs upgrade to PHP 8.1 ;)

andygrunwald, line 4:
```suggestion
FROM php:8.1.20-fpm-bullseye
```

andygrunwald: Done

bin/prepare-release.sh
```diff
@@ -14,16 +14,19 @@
 #   bin/prepare-release.sh v1.2.3
 #
 
+LANSUITE_GIT_URL="https://github.com/lansuite/lansuite.git"
+OUTPUT_DIR="/builds/"
+
 # Our own sha256sum function call.
 # That takes into account machines that don't have sha256sum but do have sha2 (i.e. macOS)
```
M4LuZ, line 21: I am a bit confused here.
I understand that this is intended to run within the container.
Why should we then care about the OS when we have it in our hands to satisfy the requirement by including e.g:
`RUN apt-get update && \
    apt-get install -y libdigest-sha-perl`

Or is this intended to be also callable natively on the host system?

andygrunwald: This is a good catch. It is meant to run inside the container.
In https://github.com/lansuite/lansuite/pull/663/commits/c922f4df7db63c87b50edaee3d35e66dcffe0032 I removed the OS independent version of the checksum function.

Dockerfile-Production-Release
```diff
@@ -0,0 +1,31 @@
...
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-get clean
+
+# Composer setup starts here. The zip extension is required for that.
+RUN docker-php-ext-install zip
+COPY --from=composer /usr/bin/composer /usr/local/bin/composer
+ENV COMPOSER_ALLOW_SUPERUSER 1
+
+# Copy prepare-release.sh script
+COPY ./bin/prepare-release.sh /usr/local/bin/prepare-release.sh
```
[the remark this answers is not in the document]
andygrunwald: The `COPY` command copies the local file `./bin/prepare-release.sh` (from the host machine where the container is built) into the container (into `/usr/local/bin/prepare-release.sh`).

For me, the target location (`/usr/local/bin/prepare-release.sh`) don't matter much. What is important here to notice: The copy command makes the container (to build a production release) independent from the local LanSuite checkout.

Even if you don't have a local LanSuite Development env running, you can build a production release.

Or did I miss your point?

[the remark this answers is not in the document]
M4LuZ: It's fine, I overlooked the part that we have a copy from host to container

M4LuZ (approved):

PR #683, opened 2023-07-14
Add support for performance measurement via Flame graphs

andygrunwald:
This Pull Request adds support to generate Flame Graphs (https://www.brendangregg.com/flamegraphs.html) via the xDebug trace functionality (https://xdebug.org/docs/trace)


andygrunwald:
@M4LuZ Any possibility for a review?

M4LuZ:
hmm. 
I do see the helpfulness of a performance profiling tool, but I fail to see what makes FlameGraphs the go-to-tool here.
Unless I understand this wrong this just provides a noninteractive graph?
Why not just refer to QCacheGrind?
If we do not want to require dependencies that the user must provide then to bundle [WebCacheGrind ](https://github.com/jokkedk/webgrind) or just [tracefile-analyser.php](https://raw.githubusercontent.com/xdebug/xdebug/master/contrib/tracefile-analyser.php) bundled with xdebug as these have readable output and are still somewhat maintained.
My preference would be on the former to allow interactive drill down from the initial overview

M4LuZ:
@andygrunwald : you may have missed my comment here (or should I be concerned that you are silent for quite a while? ;))

andygrunwald:
@M4LuZ 

> I fail to see what makes FlameGraphs the go-to-tool here

FlameGraphs are not the go-to-tool. It is a different representation and an alternative to things like QCacheGrind.

> Why not just refer to QCacheGrind?

I started to use FlameGraphs and just made a PR. There is no "competing" guide for QCacheGrind and I don't plan to create one.

> If we do not want to require dependencies that the user must provide then to bundle [WebCacheGrind ](https://github.com/jokkedk/webgrind)or just [tracefile-analyser.php](https://raw.githubusercontent.com/xdebug/xdebug/master/contrib/tracefile-analyser.php) bundled with xdebug as these have readable output and are still somewhat maintained.

This Pull Request doesn't introduce a new dependency. The doc says you need to install another tool locally. This is the same with cache grind tools.
I personally would avoid adding another dependency, because I think performance measurement tools should be local and not part of the product itself.

> My preference would be on the former to allow interactive drill down from the initial overview

A drill down is also possible with Flamegraphs on the SVG.

---

In the end, it is an attempt to add one performance profiling method into LanSuite.
Mainly because now and then, in an issue, performance is mentioned as a criterion but never proven.
If you prefer CacheGrind, I would love to see a Pull Request that describes how to use this with LanSuite to enable others to use it.
I played around with FlameGraphs, found them useful and added my efforts as an PR for the community.

If you think this is not useful, I don't mind if this is getting not merged.
Personally, I see value in this guide, as long as there is no alternative available.
Mainly because I think something in this direction is better than nothing.

M4LuZ:
Hi @andygrunwald ,

I wasn't trying to say that this is not useful.
I was under the assumption that FlameGraphs "just" creates a static image without any way to further drill down. 
Thus mentioning the alternatives where I did already know that do this to pitch the idea for that.

But as I understand I am mistaken here (even with spending due diligence by checking the project page).
In that case that solves my only issue and I'm happy to merge as-is.

M4LuZ (approved):

andygrunwald:
> I was under the assumption that FlameGraphs "just" creates a static image without any way to further drill down.

It is a SVG, which can be zoomed a lot.

Issue #596, opened 2023-05-25
Double check file includes if this are vulnerable to Path / Directory traversal


andygrunwald:
## Current Behavior

Some `include()` statements contain user-controlled input like `$_GET` variables.
This might open up the system for Path / Directory traversal.

## Expected Behavior

Ensure the include statements are save to executed only with sanitized values.

## Possible Solution

To be worked out

## Steps to Reproduce (for bugs)

See lines like

https://github.com/lansuite/lansuite/blob/f11dae125714907bef383ae5c71859eafd38278c/index_module.inc.php#L76

https://github.com/lansuite/lansuite/blob/f11dae125714907bef383ae5c71859eafd38278c/index_module.inc.php#L84

https://github.com/lansuite/lansuite/blob/f11dae125714907bef383ae5c71859eafd38278c/index_module.inc.php#L90

and related

## Context

This popped up in https://github.com/lansuite/lansuite/pull/576#discussion_r1205613974

Issue #572, opened 2023-05-15
Proposal: Using Symfony Filesystem Component for platform-independent utilities for filesystem operations


andygrunwald:
This is a proposal on introducing a new 3rd party code component into LANSuite: **Symfony Filesystem Component**

## Current Behavior

LanSuite deals with various files:
* Config files
* Cache files
* Script files for includes, etc.

Dealing means reading and writing (not only cache files, but also config files).
Pathes are often put manually together by assuming we run on a Linux system.

## Expected Behavior

A unified solution to deal with the Filesystem that properly understands the platform-dependent needs.

## Possible Solution

> The Filesystem component provides platform-independent utilities for filesystem operations and for file/directory paths manipulation.

## Positive side effects

* Platform independence
* [Converting Absolute/Relative Paths](https://symfony.com/doc/current/components/filesystem.html#converting-absolute-relative-paths)
* [Finding Directories/Root Directories](https://symfony.com/doc/current/components/filesystem.html#finding-directories-root-directories)
* Functions to avoid path traversal

## References

* [The Filesystem Component documentation](https://symfony.com/doc/current/components/filesystem.html)
* [symfony/filesystem @ GitHub](https://github.com/symfony/filesystem)

PR #710, opened 2023-08-20
index_module: Protect module inclusion for path traversal

andygrunwald:
### What is this PR doing?

We introduce a new function BuildModuleFilePath, that helps
to protect and build the right path to include a particular file
from a module.

The symfony/filesystem helps us with this by abstracting and normalizing
cross-platform path functionality.

Additionally, we switched the database statements to prepared statements.

Fixes https://github.com/lansuite/lansuite/issues/596
Fixes https://github.com/lansuite/lansuite/issues/572

### Checklist

- [x] `CHANGELOG.md` entry -> https://github.com/lansuite/lansuite/pull/708 need merge first
- [X] Documentation update -> None

### DIFF
--- a/inc/Functions/BuildModuleFilePath.php
+++ b/inc/Functions/BuildModuleFilePath.php
@@ -1,0 +1,37 @@
+<?php
+
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Filesystem\Path;
+
+ /**
+  * Builds the correct path to include a file from a particular module.
+  *
+  * $module and $file should only contain letters, numbers and valid chars like _ or -.
+  * If this is not the case, an exception will be thrown.
+  *
+  * This function also protects for path traversal attacks.
+  */
+function BuildModuleFilePath(Filesystem $filesystem, string $rootDirectory, string $module, string $file): string
+{
+    // Some files contain a `_` or `-` char.
+    // To avoid a regular expression, we allow the chars by replacing.
+    $validChars = array('-', '_');
+
+    // Check that they only contain valid chars like letters and numbers
+    if (!ctype_alnum(str_replace($validChars, '', $module)) || !ctype_alnum(str_replace($validChars, '', $file))) {
+        throw new \Exception("Path contains unexpected characters");
+    }
+
+    $pathToInclude = $rootDirectory . 'modules' . DIRECTORY_SEPARATOR . $module . DIRECTORY_SEPARATOR . $file . '.php';
+    $pathToInclude = Path::canonicalize($pathToInclude);
+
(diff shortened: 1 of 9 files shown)


andygrunwald:
@M4LuZ Branch updated and changelog entry added. Ready for review.

index_module.inc.php
```diff
@@ -77,38 +77,62 @@
...
+            $actionParameter = $request->query->get('action') ?? '';
             // 1) Search $_GET['action'] in DB (field "action")
-            $menu = $db->qry_first("SELECT file, requirement FROM %prefix%menu WHERE (module = %string%) and (action = %string%)", $modParameter, $actionParameter);
+            $sqlQuery = 'SELECT `file`, `requirement` FROM `%prefix%menu` WHERE `module` = ? AND `action` = ?';
+            $menu = $database->queryWithOnlyFirstRow($sqlQuery, [$modParameter, $actionParameter]);
             if ($menu && $menu['file'] != '') {
+                // Case like
+                //  - /?mod=info2&action=change
                 if ($authentication->authorized($menu['requirement'])) {
-                    // TODO Fix security issue
```
M4LuZ, line 85: I suspect that this was not actually a security issue as `$menu['file']` can only be modified by somebody with administrative privileges to modify the menu entries.
And that group can include any file they want anyways, so this would not be a privilege escalation.

I suspect that the gallery and forum are places where this could become more...interesting, as we have file uploads (and access) by otherwise unprivileged users

inc/Functions/BuildModuleFilePath.php
```diff
@@ -0,0 +1,37 @@
...
+  *
+  * This function also protects for path traversal attacks.
+  */
+function BuildModuleFilePath(Filesystem $filesystem, string $rootDirectory, string $module, string $file): string
+{
+    // Some files contain a `_` or `-` char.
+    // To avoid a regular expression, we allow the chars by replacing.
+    $validChars = array('-', '_');
+
+    // Check that they only contain valid chars like letters and numbers
```
M4LuZ, line 20: We may also want to prevent access to certain files (or potentially even file types).
Prevent Writes
* .htaccess
* user.ini

Prevent Reads and Writes
* config.php
* .git-credentials

Would it be a good idea to add handling for that?

index_module.inc.php
andygrunwald: `$modParameter` is still a user input (and partially sanitized due to the database query).
However, having the same logic in every include scenario doesn't hurt and makes it safe anyway.

Do you see any disadvantage to do this?

andygrunwald:
@M4LuZ Any update?

M4LuZ (approved):

PR #758, opened 2023-10-06
Module: New module `birthday`

andygrunwald:
### What is this PR doing?

This PR introduces a basic new module to show users birthday's.

The box for the navigation looks like:
Screenshot 2023-10-06 at 19 31 08

The content under http://domain/index.php?mod=birthday looks like

Screenshot 2023-10-06 at 19 31 59

This is a nice gimmick and can be used to celebrate on lan parties

None

### Checklist

- [X] `CHANGELOG.md` entry
- [X] Documentation update - Not needed - Standard module


M4LuZ:
In general I like the idea, buuuut:
This does not ensure that the users actually consent to having the information public if and when they have their birthday and how old they are. Which some could take an issue with.
What do you think about adding an option to the user details (similar to "show my adress on the map") that governs if a birthday is listed or not?

Given that it is configured to be intranet-only it _should_ be fine, but this is one of those cases that makes it clear that we still have some catching up to do to be GDPR-compliant

andygrunwald:
> What do you think about adding an option to the user details (similar to "show my adress on the map") that governs if a birthday is listed or not?

That is a great idea! Let me work on this.

> Given that it is configured to be intranet-only it should be fine, but this is one of those cases that makes it clear that we still have some catching up to do to be GDPR-compliant

I think if we have the consent of the user, this can even be an "internet" module, because most lan parties need internet access anyway.

andygrunwald:
@M4LuZ Have a look at https://github.com/lansuite/lansuite/pull/758/commits/fd2eddc4eeb91734ad039a6d6f798ca277a7ed14 and https://github.com/lansuite/lansuite/pull/758/commits/e7e2b54c0fac2249b88ffff79125db05682a3d7d

I added a new database field `show_birthday`. This can be activated/deactivated in the user settings. Based on this the user can control if the birthday is shown or not:

Screenshot 2023-10-14 at 19 31 44

andygrunwald:
@M4LuZ Any chance to get a review?

M4LuZ (approved):

PR #787, opened 2023-11-10
[Tasks] New module to manage basic tasks

andygrunwald:
### What is this PR doing?

This Pull Requests adds a basic module to manage and assign tasks.
Tasks can be used to assign responsibilities on a LAN Party.

This Module is by far not perfect and has a lot of code duplication.
This module is a basic back port until it gets working with the current LANSuite version.
I found this module while upgrading an older LANSuite version and want to play this back to upstrea/github.

Once merged, this needs a bit more work to be very useful. However, the basic functionality is there and working.

Some screenshots:

Screenshot 2023-11-10 at 19 26 49
Screenshot 2023-11-10 at 19 26 57
Screenshot 2023-11-10 at 19 27 05
Screenshot 2023-11-10 at 19 27 11

None

### Checklist

- [X] `CHANGELOG.md` entry


M4LuZ:
Hi @andygrunwald,

As far as I can see this for users to raise tickets that are then to be handled by the team, right?
Or for the team to raise and solve internally.

But that appears to be very similar if not the same to the existing {{troubleticket}} module?
Personal feeling would be to extend (and rename to a more generic term) the troubleticket-module if needed and to leave this out.
Especially if it still needs a good amount of work to be really useful.

But I got no strong feeling either way, also fine to merge if somebody wants / need it or it fulfils a purpose that the other module  can't

PR #794, opened 2023-11-12
[Documentation] Rework Upgrade Guide and add Section "Upgrade from LanSuite v4.2 to v5.0"

andygrunwald:
### What is this PR doing?

I recently upgraded an older LanSuite instance.

For this, I had first-hand experience on what steps are necessary to upgrade.

This Pull Request reworks the Upgrade Guide a bit:

- Add specific instructions on v4.2 to v5.0
- Made it a bit more concise (e.g. i think it is not our responsibility to inform the admin if he/she should inform their users)
- Removed the git master option - IMO, this should only be done by users who know what they are doing

None

### Checklist

- [X] Documentation update


M4LuZ:
@mrflobow contacted me because he apparently had issues in updating his dev environment while following the guide.
(which indicates that some more work is apparently needed)
I've asked him to post the issues and questions here, so it can be added.

M4LuZ:
While we wait:

> Made it a bit more concise (e.g. i think it is not our responsibility to inform the admin if he/she should inform their users)

The guide lists all steps that should be done to avoid issues before/during/after the upgrade.
And lack of communication / planning is one very often seen.

So I would leave it in. 
I'm confident that readers can figure out for themselves if they need to do that or not

mrflobow:
Hello Andy and Maluz

I would like to get the DevSetup running I tried the steps on the offical instructions but I'm not able to get it running.


My comments

DEV Setup?

```
git clone https://github.com/lansuite/lansuite.git
cd lansuite
# Erstelle eine Konfigurations Datei in ./inc/base/config.php , eine Beispiel Konfiguration kannst du unten entnehmen
touch ./inc/base/config.php
# > Not sure if applicable for dev setup
chmod 0777 ./inc/base/config.php
# > Would case 10000 commit files.
chmod -R 0777 ./ext_inc/
# Added step to get mysql setting
cp docker/mysql.env.dist docker/mysql.env
#> Image must be build otherwise its taking an image 3years old with php7.x
docker build . -t lansuite/lansuite --platform linux/amd64
#> Add -d for background
docker-compose up -d
docker-compose run php composer install

# Add
# Page is reachable add http://localhost:8080 
```

> Sample should match with mysql.env for lazy devs :D

```
[lansuite]
version=Nightly
default_design=simple
chmod_dir=777
chmod_file=666
debugmode=0

[database]
server=mysql
user=lansuite
passwd=dev1234
database=lansuite
prefix=ls_
charset=utf8
```

1.
Bildschirmfoto 2023-11-18 um 12 32 57


2.
Bildschirmfoto 2023-11-18 um 12 33 28

3.
Bildschirmfoto 2023-11-18 um 12 34 25

Not sure what todo next.

website/docs/guides/guides-upgrade.md
```diff
@@ -45,77 +45,124 @@ If you executed the previous step already on your server, then obviously nothing
...
+```
+
+### Configuration: `database.charset`
+
+Add the following line to your configuration at `/inc/base/config.php`:
+
+```
+[...]
+[database]
+charset = "utf8mb4"
```
M4LuZ, line 133: There was a discussion in the past about what the default value should be.
`utf8mb4` is set by default, but if I'm not mistaken that causes display issues on old installations where content is stored as `latin1`. Unless I missed that being fixed in the meantime

website/docs/guides/guides-upgrade.md
```diff
@@ -45,77 +45,124 @@ If you executed the previous step already on your server, then obviously nothing
...
+[...]
+```
+
+### Fonts in `ext_inc/pdf_fonts`
+
+Delete the following files
```
M4LuZ, line 139: I feel that we can provide a small script here to skip most if not all of the manual work

website/docs/guides/guides-upgrade.md
```diff
@@ -45,77 +45,124 @@ If you executed the previous step already on your server, then obviously nothing
...
+
 * The base configuration file from `inc/base/config.php`
 * Anything under `ext_inc`
 * Custom designs from `designs/`
 
 Copy them to the same place on the new installation, overwriting everything there.
 
-### Check file access rights
-
-Depending on the user you did the installation with it may be required to reset the file and folder access rights.
```
M4LuZ, line 80: What is the rationale of removing this?
Unless file operations are done in scope of the webserver user, ownership may be wrong, so may be access rights

website/docs/guides/guides-upgrade.md
```diff
@@ -45,77 +45,124 @@ If you executed the previous step already on your server, then obviously nothing
...
 ### Run specific release tasks
 
 It may be required to run additional steps to prepare everything for the new version.
-These will be detailed in either `README.md` or `UPGRADE.md`
+Please check the upgrade guide to your specific version below.
+
+### Execute LanSuite upgrade logic
+
+Visit "Admin-Page" -> "Lansuite updaten / reparieren" ->> "Datenbank updaten und verwalten".
+This is also available at `http(s)://<your-domain>/index.php?mod=install&action=db`.
```
M4LuZ, line 77: We may need to add also something for menu updates. I noticed that changes to the `menu.xml` files are apparently not taken over automatically

website/docs/guides/guides-upgrade.md
```diff
@@ -4,39 +4,39 @@ title: Upgrade to a newer version
...
-### Plan
+### Download an official release
 
-It is recommended to tell your users (and fellow administrators) in advance that your LanSuite instance won't be 100% functional available during the upgrade.
-That should be well done in advance if you run a larger installation, for smaller ones no one may notice. But better be safe than sorry.
+1. Download the release from [GitHubs Releases page](https://github.com/lansuite/lansuite/releases)
+2. Extract the archive on your machine
 
-### Prepare Upgrade files
+### Execute a backup
```
M4LuZ, line 19: The order of operation is wrong.
The system should be locked first and then the backup created.
If you do it the other way around then there is a risk that additional changes are done that are missing from the backup.
_Especially_ if you also do not tell other admins that you are doing an upgrade ;)

website/docs/guides/guides-upgrade.md
```diff
@@ -4,39 +4,39 @@ title: Upgrade to a newer version
...
-* Your target system running PHP >= 7.0 and MySQL >= 5.6.3
+* Your target system need to run PHP >= 8.1 and MySQL >= 5.7
 * A fair bit of time
 
 ## Preparation
 
-### Plan
+### Download an official release
 
-It is recommended to tell your users (and fellow administrators) in advance that your LanSuite instance won't be 100% functional available during the upgrade.
```
M4LuZ, line 24: As per comment:
The guide lists all steps that should be done to avoid issues before/during/after the upgrade.
And lack of communication / planning is one very often seen.

So I would leave it in.
I'm confident that readers can figure out for themselves if they need to do that or not

website/docs/guides/guides-upgrade.md
```diff
@@ -45,77 +45,124 @@ If you executed the previous step already on your server, then obviously nothing
...
+* `ext_inc/pdf_fonts/.gitkeep`
+
+### Converting custom fonts
+
+If you have added your own custom fonts into `ext_inc/pdf_fonts`, you need to convert them into the new fpdf format:
 
-* Please ensure that export and import of Database images use the same character encoding. Using the same client on the same system should ensure this, but be cautious if dump and import are done on different systems/clients
-* MySQL 5.7 changes default behavior for GROUP BY clauses. This may lead to unexpected errors in some cases. See [#117](https://github.com/lansuite/lansuite/issues/117) for details
-* The master branch is not usable without pulling in additional resources via composer. You must do this first to obtain a runnable installation!
-* You have to run a database table upgrade as an extension of the IP field is required to store logging information
```
M4LuZ, line 120: Do we need to provide a conversion logic for existing DB entries in IPv4?

M4LuZ (changes requested):

website/docs/guides/guides-upgrade.md
andygrunwald: This is true - Mainly to a different charset.
I do think we should not take responsibility on converting a database for others. I think there are better articles in the internet that cover this.

Maybe adding one sentence to the doc to mark this?

website/docs/guides/guides-upgrade.md
andygrunwald: We could provide one. I personally would love to see that we don't make it a requirement for this PR to get merged. Mainly, because I believe we first should make it work and our docs are not good. Manual work works at first and moves us forward vs. providing everything in a perfect shape.

website/docs/guides/guides-upgrade.md
andygrunwald: The installation script checks the file permission and warns the user. Thats why I don't think it should be a manual step. On top of this it doesn't say
- which file permission to check
- what are the required file permission

Hence it is providing limited value.

website/docs/guides/guides-upgrade.md
andygrunwald: How do you do it? Where? Would you mind adding this step?

website/docs/guides/guides-upgrade.md
andygrunwald: I personally think we need to set boundaries here between us and the user/operator. We can't fix everything for them.

I never have seen such a note in any other system / open source upgrade path. I do think a short and concise upgrade doc that covers the requirements is read more often than a "this covers everything" doc.

Anyway. I would still remove it and focus on the system. But if you think this is useful, we can keep it in.

website/docs/guides/guides-upgrade.md
andygrunwald: This part is removed. As far a I know this is not needed anymore. I have updated my system already and never did such a conversion

website/docs/guides/guides-upgrade.md
andygrunwald: Fixed in https://github.com/lansuite/lansuite/pull/794/commits/a1a9aebebca65f92766cf496a36a0da3b311b7a2

website/docs/guides/guides-upgrade.md
andygrunwald: Fixed in https://github.com/lansuite/lansuite/pull/794/commits/4ab9c4d50b3ab8258c032c78c6dfbe534c70442e

andygrunwald:
@mrflobow I had a quick check on your journey. It seem to be relate to the MySQL database.
You start your docker setup in background via

```
#> Add -d for background
docker-compose up -d
```

Can we verify that MySQL was booted up successfully?
Or that the user

```
user=lansuite
passwd=dev1234
```

has permission to create a database?

@M4LuZ The comment by @mrflobow is related to a development installation, not on how to upgrade an existing LanSuite installation. Hence, I would suggest to open up a new ticket on this.

website/docs/guides/guides-upgrade.md
M4LuZ: It's the [debian way of doing things](https://www.debian.org/releases/stable/i386/release-notes/ch-upgrading.en.html#inform-users) ;)

website/docs/guides/guides-upgrade.md
M4LuZ: I think we discussed somewhere that we intend to cater also to not that tech-savvy users.
Thus it would be hard to expect them to figure that out for themselves.

Interestingly there were already some on the old page
https://web.archive.org/web/20090113122545/http://lansuite.orgapage.de/index.php?mod=wiki&action=show&name=Latin1toutf8
This we would just need to translate and integrate.
But yes, there should be at least a remark and the recommendation to go with `latin1` if issues occurr and/or unable to convert

website/docs/guides/guides-upgrade.md
M4LuZ: The installation script is not executed as part of the instructions.
So that does not help.
If I remember correctly it also does not tell what should be set, just that permissions are not sufficient.

I do agree with you that it would be better to detail what permissions should be set.
But that does not mean that the information that this may need to be done should be removed alltogether.

M4LuZ: Now looking deeper at the old upgrade instructions:
> In der Datei inc/base/config.php den Wert von Configured auf 0 zurücksetzen
Nun die Webseite aufrufen. Es erscheint die Installation. Diese ganz normal durchklicken und die Datenbank Zugangsdaten eingeben jedcoh NICHT DAS HÄCKCHEN setzen, dass die DB löscht. Die Installation sollte nun automatisch updaten.

Huh. If that works it covers most of what is needed.

website/docs/guides/guides-upgrade.md
M4LuZ: I'm still figuring that out myself.
Re-Running the install wizzard as marked above could be the solution.
Will get back to you if I find something, would suggest to add a TODO and not wait for us to figure that out

website/docs/guides/guides-upgrade.md
M4LuZ: Had a look. It appears that old column values are dropped with the related column change.
So there is nothing to convert once the DB update is done.
Not sure if feature or bug :/

M4LuZ:
> @M4LuZ The comment by @mrflobow is related to a development installation, not on how to upgrade an existing LanSuite installation. Hence, I would suggest to open up a new ticket on this.

That's fine by me

website/docs/guides/guides-upgrade.md
M4LuZ: I've added a bit of text in that regard. 
Feel free to revert, modify or accept

website/docs/guides/guides-upgrade.md
M4LuZ: It's not a requirement, but I feel that this could be done easily and would eliminate the chance of human error

website/docs/guides/guides-upgrade.md
andygrunwald: > I think we discussed somewhere that we intend to cater also to not that tech-savvy users.

That is a good requirement, however, I don't see this realistic in the current (and near future) state of this software. There is tons of work to do to make this happen. I won't be able to invest this amount of time.

@M4LuZ I see you did a change already: https://github.com/lansuite/lansuite/pull/794/commits/df478fba540978eeed34a02d713230ddb65be7d0

I think it is the wrong direction, however, if you like to have it this way, I am fine with it.

website/docs/guides/guides-upgrade.md
andygrunwald: I created https://github.com/lansuite/lansuite/issues/893 and won't add a TODO

website/docs/guides/guides-upgrade.md
andygrunwald: Storing the ipv4 addresses without a proper reason is not gdpr confirm either. In this case, I don't think we need to fix this.

On top of this, people have a database backup. If they really need the IPv4 addresses, there is a possibility to add those.

andygrunwald:
@M4LuZ Which items are missing in this PR?

website/docs/guides/guides-upgrade.md
M4LuZ: > There is tons of work to do to make this happen. I won't be able to invest this amount of time.

Neither will I. But in this case it was just a few minutes to add the related text.

> if you like to have it this way, I am fine with it.

So am I ;)

> Feel free to revert, modify or accept

website/docs/guides/guides-upgrade.md
M4LuZ: I would suggest to add a note that this happens and leave it with that.
log is (supposed to be) cleaned up regularly via cron2, so these should disappear in reasonable time anyways.

Ad GDPR:
The proper reason is legimimate interest and is actually fine as long as it is disclosed transparently (where we fail at the moment). 
(Apparently there are also some court cases in DE/AU where dynamic IPs were not considered personal information, but as so often it is not clear-cut)

M4LuZ:
I've resolved the conversations finished and have only these two remaining:
* Suggestion: Add a note that IP entries are removed from logs
* Test if menu-writing and permissions are checked / fixed when setting `configured=0` as suggested in legacy upgrade guide. But this can be done within #893 to get this closed

website/docs/guides/guides-upgrade.md
```diff
@@ -4,118 +4,203 @@ title: Upgrade to a newer version
...
 Either by extracting (`gunzip ls_backup.tar.gz && tar xvf ls_backup.tar`) or moving back the copied folder (`mv <ls_folder>.backup <ls_folder>`).
 Or: Re-uploading your backup, if that was the way you went.
 
-### Unlock installation
+## Pitfalls and Known Bugs
 
-As the backup was taken at a point where the installation was locked, it may still be required to unlock it
-This would be done by reverting whatever was done to facilitate the original lock
+* Please ensure that export and import of Database images use the same character encoding. Using the same client on the same system should ensure this, but be cautious if dump and import are done on different systems/clients.
+* The git `master` branch is not usable without pulling in additional resources via composer. You must do this first to obtain a runnable installation!
```
M4LuZ, line 118:
```suggestion
* IP Addresses from existing log entries are removed due to a change of the column format for IPv6 support
```

M4LuZ (approved):

website/docs/guides/guides-upgrade.md
andygrunwald: Done in https://github.com/lansuite/lansuite/pull/794/commits/929e27a9c2898c8c2fbe7ea9ad769069dbffbd91

mrflobow:
@andygrunwald tried it also in a vm with all service installed . The issue is the use of old db class and new which leads to fail in the wizard . Yes this is unrelated to upgrade , the inital install fails neither its docker or vm.

So may try to reproduce on a vanilla vm like me . At the moment have not the time to invest further.

PR #795, opened 2023-11-12
Define minimum PHP and MySQL Versions as PHP Constants

andygrunwald:
### What is this PR doing?

This PR moves the minimum PHP and MySQL versions into PHP constants.

Side note: I think we should remove official MariaDB support. Mainly because it is not a drop in replacement for MySQL, but rather have some in-place capability to run like MySQL. Maria DB moves away from the drop-in more and more. This comment should not block this PR.

None

### Checklist

### DIFF
--- a/inc/base/define.php
+++ b/inc/base/define.php
@@ -5,6 +5,9 @@
  */
 define('LANSUITE_VERSION', '5.0-dev');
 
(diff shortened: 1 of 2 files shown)


M4LuZ (approved):
Just a thought:
We should be able to define a single point of truth for versioning and fetch/insert these during the build process, shouldn't we?

PR #798, opened 2023-11-14
Fix SQL error in Cashmgr and Mastersearch2 when two dropdown search fields are involved

andygrunwald:
### What is this PR doing?

In the CashManager module, two dropdowns are used for searching.
The MasterSearch class generates a SQL `HAVING` statement based on this.
The field that appears in the `HAVING` statement is not appearing in the `SELECT` statement of this query.

This throws an SQL error, because the documentation at https://dev.mysql.com/doc/refman/8.0/en/select.html says:

> The SQL standard requires that HAVING must reference only columns in the GROUP BY clause or columns used in aggregate functions. However, MySQL supports an extension to this behavior, and permits HAVING to refer to columns in the [SELECT](https://dev.mysql.com/doc/refman/8.0/en/select.html) list and columns in outer subqueries as well.

### PR Dependency

https://github.com/lansuite/lansuite/pull/797 needs to be merged before.

None

### Checklist

### DIFF
--- a/modules/cashmgr/myaccounting.php
+++ b/modules/cashmgr/myaccounting.php
@@ -1,8 +1,9 @@
 <?php
 
-if ($_GET['act'] == "him" && $auth['type'] < \LS_AUTH_TYPE_SUPERADMIN) {
+$actParameter = $_GET['act'] ?? '';
+if ($actParameter == "him" && $auth['type'] < \LS_AUTH_TYPE_SUPERADMIN) {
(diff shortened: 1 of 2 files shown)


andygrunwald:
@M4LuZ Are we able to merge this soon? I would like to continue with fixing and want to avoid piling up PRs on each other.

M4LuZ (approved):

PR #863, opened 2024-01-10
Fix not working board delete

### LINKED ISSUE #861: Masterdelete / Board: Forum deletion fails without error
## Expected Behavior

When clicking the icon to delete a Board it should either remove it or at least report _somewhere_ that it had an issue

## Current Behavior

clicking the related icon just shows the overview page with the board still there.
No error displayed, nothing in the log.
Normally a confirmation display should be shown.

Deleting threads and comments works as expected.

## Possible Solution

1. See if this is an issue with the call from `board/delete.php` and fix it
2. See if this is an issue with the `MasterDelete` class and fix it
3. Replace delegation to `MasterDelete` with native code as part of the class or separate function

1. Create a board
2. Click on the deletion icon

## Context

I found this while validating functionality after refactoring `board/thread.php` to use the new db class

## Your Environment

* Version used: Git master
* Operating System and version: Win 10, PHP 8.2, MySQL 8.0
* Enabled features:

M4LuZ:
### What is this PR doing?

adds implit typecasting to int for match statement to hit cases `2` and `10`
This failed previously due to `$stepParameter` being a string and thus always the default case to display the board was exected

Fixes #861 

### Checklist

- [x] `CHANGELOG.md` entry
- [x] ~Documentation update~ (bugfix only)

### DIFF
--- a/modules/board/delete.php
+++ b/modules/board/delete.php
@@ -20,7 +20,7 @@ if ($pidParameter) {
 
 // Delete board
 } else {
-    match ($request->query->getInt('step',0)) {
+    match ($stepParameter) {
         2 => $md->Delete('board_forums', 'fid', $fidParameter),
         10 => $md->MultiDelete('board_forums', 'fid'),
         default => include_once 'modules/board/show.php'
--- a/modules/board/delete.php
+++ b/modules/board/delete.php
@@ -2,23 +2,27 @@
 
 $md = new \LanSuite\MasterDelete();
 
+$pidParameter = $request->query->getInt('pid', 0);
+$tidParameter = $request->query->getInt('tid', 0);
+$fidParameter = $request->query->getInt('fid', 0);
+$stepParameter = $request->query->getInt('step', 0);
+
+
 // Delete post
-$pidParameter = $_GET['pid'] ?? 0;
-$tidParameter = $_GET['tid'] ?? 0;
-$stepParameter = $_GET['step'] ?? 0;
+
 if ($pidParameter) {
     $md->DeleteIfEmpty['board_threads'] = 'tid';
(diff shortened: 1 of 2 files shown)


modules/board/delete.php
```diff
@@ -16,7 +16,7 @@
 
 // Delete board
 } else {
-    match ($stepParameter) {
+    match ((int)$stepParameter) {
```
andygrunwald, line 19: What do you think about adding
```
$stepParameter = $_GET['step'] ?? 0;
$stepParameter = (int) $stepParameter;
```

on top of the file? This would ensure the parameter is always an integer and not only in this particular match case. I can't think of a scenario, where a string would be beneficial here.

andygrunwald (changes requested):

modules/board/delete.php
M4LuZ: Now that you forced me to think about it again:
I would say the best way would be to use `$request->query->getInt('step',0)`.
As that enforces a default value, does the type casting and refactors already to utilize functionality of HttpFoundation.
I would be so lazy to drop that directly into the match statement, but I guess the cleaner way would be to initialize a variable on top. 
And yes, all of those step variables are integers...which is also we may want to change or at least define as constants

M4LuZ:
Hi Andy.,

I am aware that there are some cases where values are written back to the superglobals(most prominently the escaping of values also in index.php), but this is not one of them.
So my thought is that this can be replaced bit by bit, as long as we are aware of these fringe cases and solve them before touching the code where values are accessed.

andygrunwald:
@M4LuZ If we do it bit by bit, I suggest we focus on the writes first because those are mostly the lines that produce the bugs.

I am good with merging it. Thanks for finding it!

andygrunwald (approved):

PR #870, opened 2024-01-11
First draft of generic file accessor class

M4LuZ:
### What is this PR doing?

WIP Implementation of a secure file accessor that makes working with files and folders easier while simultaneously preventing directory traversal both below LS root and any set intermediate path.
All current file accesses to be replaced with this

no issue created

### Checklist

- [ ] `CHANGELOG.md` entry
- [ ] Documentation update

### DIFF
--- a/inc/Classes/File.php
+++ b/inc/Classes/File.php
@@ -1,0 +1,96 @@
+<?php
+
+namespace LanSuite;
+
+/**
+ * Class File
+ * Generic class for file accessor functions
+ */
+
+class File
+{
+    private static bool $_BASEPATH_INITIALISED = false;
+    private static string $_LS_BASEPATH;
+    private string $_intermediatePath;
+
+    public function __construct()
+    {
+        //Ensure basepath is set on first instantiation
+        if (!self::$_BASEPATH_INITIALISED) {
+            self::$_LS_BASEPATH = realpath(__DIR__ . '..'. DIRECTORY_SEPARATOR . '..'); //move down from /inc/classes/ to LS root
+            self::$_BASEPATH_INITIALISED = true;
+        }
+    }
+
+    /**
+     * Builds a file path with the relative path and checks for path traversals
+     * 
+     * @var    string $relativePath the relative path to be accessed
+     * @return string|bool The full path or false if attempted directory traversal
+     */
+    public function getFullPath($relativePath) :string|bool 
+    {
+
+        $path = realpath(self::$_LS_BASEPATH . $this->_intermediatePath . $relativePath);
+        //ensure that resulting path is still below the base path, return false if not
+        if (str_starts_with($path, self::$_LS_BASEPATH . $this->_intermediatePath)) {
+            return $path;
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * Sets a relative path on top of the basepath for any following operation
+     * 
+     * @var    string $intermediatePath a path to be added on top of basepath
+     * @return bool true if set, false if not secure
+     */
+    public function setIntermediatePath($intermediatePath) 
+    {
+        //ensure path ends with system directory separator to prevent directory change
+        if (!str_ends_with($intermediatePath, DIRECTORY_SEPARATOR)) {
+            $intermediatePath .= DIRECTORY_SEPARATOR;
+        }
+        //ensure new path is not below basepath
+        $path = realpath(self::$_LS_BASEPATH . $intermediatePath);
+        if (str_starts_with($path, self::$_LS_BASEPATH)) {
+            $this->_intermediatePath = $intermediatePath;
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * Returns relative path 
+     * 
+     * @return string the combination of base + intermediate path
+     */
+    public function getRelativePath() 
+    {
+        return self::$_LS_BASEPATH . $this->_intermediatePath;
+    }    
+
+    /**
+     * Returns value of relative path
+     * 
+     * @return string the intermediate path set on top of the base path
+     */
+    public function getIntermediatePath() 
+    {
+        return $this->_intermediatePath;
+    }
+
+    /**
+     * Returns value of base path
+     * 
+     * @return string the base path set on top of the base path
+     */
+    public function getBasePath() 
+    {
+        return self::$_LS_BASEPATH;
+    }
+
+
+}


M4LuZ:
@andygrunwald : Would appreciate your feedback on the general design and how that works based on the example file changed

inc/Classes/File.php
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Class File
+ * Generic class for file accessor functions
+ */
+
+class File
+{
+    private static string $_basePath='';
+    private string $_relativePath='';
+    private Filesystem $_fileSystem;
```
andygrunwald, line 16: I am personally not a fan of the `$_*` way, because the `private` keyword already marks it as a private variable. Any rational behind it?

inc/Classes/File.php
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * Constructor sets base path and initializes FS object
+     * 
+     * @param Filesystem $fs Filesystem-object for dependency injection
+     */
+    public function __construct(Filesystem $fs = new Filesystem)
+    {
+        $this->_fileSystem = $fs;
+        self::$_basePath = Path::canonicalize(ROOT_DIRECTORY);
```
andygrunwald, line 26: Can we also inject `ROOT_DIRECTORY` as a function default argument? This makes unit testing easier.

inc/Classes/File.php
```diff
@@ -0,0 +1,129 @@
...
+     */
+    public function getFullPath($path) :string|bool 
+    {
+
+        $path = Path::canonicalize(self::$_basePath . $this->_relativePath . $path);
+        //ensure that resulting path is still below the base path, return false if not
+        if (str_starts_with($path, self::$_basePath . $this->_relativePath)) {
+            return $path;
+        } else {
+            return false;
```
andygrunwald, line 46: What do you think about throwing an exception rather returning a boolean?
This would make the return value be always a string.

Alternative would be an empty string.

inc/Classes/File.php
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    /**
+     * Sets a relative path on top of the basepath for any following operation
+     * 
+     * @param  string $relativePath a path to be added on top of basepath
+     * @return bool true if set, false if not secure
+     */
+    public function setRelativePath($relativePath)
```
andygrunwald, line 56: Definition of a return value as bool would be nice.

inc/Classes/File.php
```diff
@@ -0,0 +1,129 @@
...
+    /**
+     * outputs file content when access OK
+     * 
+     * @param string $filePath The relative path to the file to be output
+     */
+    public function outputFileContent(string $filePath): void
+    {
+        $fullPath = $this->getFullPath($filePath);
+        if ($fullPath && $this->_fileSystem->exists($fullPath)) {
+            readfile($fullPath);
```
andygrunwald, line 77: This function would have the side effect of outputting it directly to the output buffer and has a side effect with it.

What is the rational behind it?
Wouldn't it be better to return the file content and leave the caller the option to output it?

inc/Classes/File.php
```diff
@@ -0,0 +1,129 @@
...
+        }
+        return;
```
andygrunwald, line 79: This is not needed as the methods ends already.

inc/Classes/File.php
```diff
@@ -0,0 +1,129 @@
...
+     * @return bool true if accessible, false if not
+     */
+
+    public function exists(string $filePath): bool
+    {
+        $fullPath = $this->getFullPath($filePath);
+        if ($fullPath) {
+            return $this->_fileSystem->exists($fullPath);
+        }
+        return false;
```
andygrunwald, line 95: If `getFullPath` always returns a empty string, we can switch this to

```suggestion
        return $this->_fileSystem->exists($fullPath);
```

andygrunwald (changes requested):
I am not 100% sure about the design of the class. Or lets say, purpose of the class.

Is the main purpose to represent one single file? Or is the purpose to enable Lansuite to initiate a File() instance once and handle multiple files?

From the name of the class, my first thought would be to handle one single file. If we loop over multiple files, for each file, a `File()` instance would be created.

In this case, I suggest
* to add the file path for the constructor as a first parameter
* to remove the argument from functions like `getFullPath` and `exists`, because the class only handles one file

In this case, it would also make sense to make a "delete" and "output" method.

LanSuite can still provide a helper function like "NewFile()" which sets the filesystem injection and root path.

Overall, I think we should do this, because we have multiple places where we deal with files.

If you think this is useful, I can also iterate over it a bit if you want :)

M4LuZ:
HI @andygrunwald ,

thanks for the helfpul feedback.

Basically this was inteded as helper class for all file access operations.
Eplicitly not on a per-file basis, but more as per-file-access-context.
So that it can be used to define a subpart of the filesystem as area where access can/will hapen, potentially providing also further scoping options (e.g. filename filtering, extension / MIME-type filtering, both with a default set and the option to set additionals per object instance).

I am pondering whether addidng direct access methods like `outputFileContent` makes sense to it.
The intent was to ensure with that that the file path is OK before the file is read, thus  that method .
This could be potentially be moved to a separate class for direct file operations?

I was close to naming the Class `FileAccessor`, based on your comments I think that helps to understand purpose and usage a lot better.

andygrunwald:
What do you think about a pure Object Oriented approach here?
Creating two classes:
- `File`
- `FileCollection`

### `File`

It is a single file and offers functions on a single file. Things like

- Exists
- Output
- Rename
- Delete
- ...

### `FileCollection`
Represents a list/number of files. Can be initialized via

- An array of files
- By a folder

Internally, it has multiple `File()` objects.

## Why?

This way we would have a clean and unit testable result. Responsibilities are clear and the security aspects (which triggered the PR) can also be taken into account.
The `FileCollection` can be used in modules like gallery, downloads, etc.

Thoughts?

inc/Classes/File.php
M4LuZ: Yes and maybe.
For unit testing purposes it makes sense, from the intent of the class it may not.
As that is to ensure that no file access happens below the root directory in all use cases.
If the root directory can be modified with the intialisation, then that has the risk of improper use.

inc/Classes/File.php
M4LuZ: It is worth to to raise the generic development question if we want to change exception handling from return values to exceptions.
I'm fine with that, just following (for now) the generic codestyle that the rest has

inc/Classes/File.php
M4LuZ: ah yes, missed that.
My copout is that this is the first draft for discussion ;)

inc/Classes/File.php
andygrunwald: Why not provding a `public static` function that does set the respective values?
This way, this can be used in both ways.

inc/Classes/File.php
andygrunwald: I think it is better to keep a consistent return type rather than mixed return types. PHP allows this, but PHP is also moving into more and more type safe variations. Why also not make use of it?

inc/Classes/File.php
M4LuZ: This was added due to the usage of readfile in `modules/picgallery/download.php`.
As per generic comment:
The intent was to ensure that the file path is OK before the file is read, thus that method .
This could be potentially be moved to a separate class for direct file operations?

inc/Classes/File.php
M4LuZ: True that!

inc/Classes/File.php
M4LuZ: I see a point that it allows to identify private attributes purely on the name without looking up the defintion, but also do agree that access within the code is not as nice as with just camelcase naming.

inc/Classes/File.php
M4LuZ: As said: That is a generic development question, I'm fine either way

inc/Classes/File.php
M4LuZ: that has the risk of being used to set a different basepath for one object instance, but affecting all of them instead.
Unless we ensure that basepath is only set once during intialization and cannot be modified after that (actually did that in the first iteration and had a static bool `basePathInitialised` for that).
Fine for me to go that way (...again)

M4LuZ:
> Thoughts?

#worksforme ;)

Was mentally already halfway there, your idea of using an array of files for initialisation for FileCollection makes a ton of sense.
For most of the proposed `File` class methods I would use functions that ``Symphony/Filesystem` already provides, either as internal calls or maybe even by inheriting them as a child class?

andygrunwald:
> For most of the proposed File class methods I would use functions that Symphony/Filesystem already provides

This is true. Still, I would shim them away from the actual implementation and create a LanSuite File class as a Wrapper. This way we can exchange the lib as we want, without doing a big rewrite.

inc/Classes/File.php
M4LuZ: Underscores have been removed

M4LuZ:
@andygrunwald: I'll still rework a bit of code in `modules/downloads/show.php`, but this should be fine for (final) review otherwise

PR #887, opened 2024-02-05
introduce separate API keys for maps, translate and analytics

M4LuZ:
### What is this PR doing?

This splits off the currently single API key definition into three dedicated values for translation, maps and analytics API
This prevents the permanent (and propably non-GDRP-compliant) inclusion of google analytics if an API key is defined for use in gmaps

### Checklist

- [x] `CHANGELOG.md` entry
- [x] Documentation update

### DIFF
--- a/modules/install/mod_settings/config.xml
+++ b/modules/install/mod_settings/config.xml
@@ -421,13 +421,6 @@ nospamfor.us</default>
         <default></default>
         <description>Google API-Key um Google-Maps Code zu aktivieren</description>
       </item>
-        </item>
-        <item>
-        <name>google_translate_key</name>
-        <type>string</type>
-        <default></default>
-        <description>Google API-Key um Google-Translate Code zu aktivieren</description>
-      </item>
     </items>
   </group>
   <group>
--- a/modules/install/mod_settings/config.xml
+++ b/modules/install/mod_settings/config.xml
@@ -413,7 +413,20 @@ nospamfor.us</default>
         <name>google_analytics_id</name>
         <type>string</type>
         <default></default>
-        <description>Google-Analytics-ID. ID um Google-Analytics Code zu aktivieren</description>
+        <description>Google API-Key um Google-Analytics Code zu aktivieren</description>
+      </item>
+        <item>
+        <name>google_maps_key</name>
+        <type>string</type>
+        <default></default>
+        <description>Google API-Key um Google-Maps Code zu aktivieren</description>
+      </item>
+        </item>
+        <item>
+        <name>google_translate_key</name>
+        <type>string</type>
+        <default></default>
+        <description>Google API-Key um Google-Translate Code zu aktivieren</description>
       </item>
     </items>
   </group>
--- a/website/docs/getting-started/settings.md
+++ b/website/docs/getting-started/settings.md
@@ -10,7 +10,7 @@ Below you find an explanation of the various config settings you find in the Lan
 
 ## Common settings
 
-To be added
+To be added (TODO)
 
 ### Website-specific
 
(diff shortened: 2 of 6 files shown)


modules/install/mod_settings/config.xml
```diff
@@ -413,7 +413,20 @@ nospamfor.us</default>
...
+      </item>
+        <item>
+        <name>google_maps_key</name>
+        <type>string</type>
+        <default></default>
+        <description>Google API-Key um Google-Maps Code zu aktivieren</description>
+      </item>
+        </item>
+        <item>
+        <name>google_translate_key</name>
```
andygrunwald, line 426: Where is this used? This seem to be added, but not used.

andygrunwald (changes requested):
I like this PR and would like to see it in the master branch.

I also like that you have added a new setting documentation.
What is still missing: A note in the upgrade guide.

I suggest you add a commit into https://github.com/lansuite/lansuite/pull/794 OR we merge https://github.com/lansuite/lansuite/pull/794 and rebase this PR to add this setting (because this might be a breaking change switching google API keys)

modules/install/mod_settings/config.xml
M4LuZ: Yes. related translation code is defunct (see #842) and will need an API key to work again.
I thought to add that already, but it can also be left out and included in the PR to fix that

website/docs/getting-started/settings.md
```diff
@@ -0,0 +1,25 @@
+---
+id: settings
+title: Settings
+sidebar_position: 4
+---
+
+# LanSuite generic config settings
+
+Below you find an explanation of the various config settings you find in the LanSuite Admin Area
```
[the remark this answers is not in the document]
M4LuZ: I'm with you on the duplication of "find", but to my knowledge this is otherwise (more) correct as-is.

"will" is not correct, unless that is meant in regards to the documentation not being complete _yet_. 
But there is documentation to be found when the user reads this, thus technically present tense is correct.

"encounter" is normally used in scope of an unexpected situation that occurrs. 
But fine to replace that with e.g. "see", to prevent duplicate use of find.
Or just remove it alltogether: "config settings in the LanSuite Admin Area"

Interpunctuation is not really needed, as there is no following sentence to separate from (but no issue to add that)

modules/install/mod_settings/config.xml
andygrunwald: I prefer to bundle the translation API key in the related change on #842.

website/docs/getting-started/settings.md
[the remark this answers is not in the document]
andygrunwald: I use grammarly.com for a lot of English communication, and mostly, it has a better "readable" result.
However, I am fine as it is.

modules/install/mod_settings/config.xml
M4LuZ: Removed together with the related documentation.
There was a duplication of `</item>`, so it was good to touch that file again anyways

M4LuZ:
> What is still missing: A note in the upgrade guide.
> 
> I suggest you add a commit into https://github.com/lansuite/lansuite/pull/794 OR we merge https://github.com/lansuite/lansuite/pull/794 and rebase this PR to add this setting (because this might be a breaking change switching google API keys)

I've held back a review of #794 due to the pending feedback for the issues of @mrflobow, but I guess that was not related? 
Will have a look at that then.

M4LuZ:
I've added a documentation block to #794 as I think(/hope) this will be merged before the upgrade doc is ready

andygrunwald (approved):

PR #891, opened 2024-02-10
[tournament2] Fix PHP warnings like undefined array key or access array offset on type bool

andygrunwald:
### What is this PR doing?

[tournament2] Fix PHP warnings like undefined array key or access array offset on type bool

### Merge order

This PR depends on https://github.com/lansuite/lansuite/pull/890
https://github.com/lansuite/lansuite/pull/890 needs to be merged before.

None

### Checklist

### DIFF
--- a/modules/tournament2/join.php
+++ b/modules/tournament2/join.php
@@ -49,8 +49,9 @@ if ($tteam->SignonCheck($tournamentid)) {
                 $error = array();
 
                 // If joining an existing team
-                if (($_POST['existing_team_name'] != "") and ($tournament['teamplayer'] > 1)) {
-                    $success = $tteam->join($_POST["existing_team_name"], $auth["userid"], $_POST["password"]);
+                $existingTeamNameParameter = $_POST['existing_team_name'] ?? '';
+                if (($existingTeamNameParameter != "") && ($tournament['teamplayer'] > 1)) {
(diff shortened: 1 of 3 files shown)


modules/tournament2/join.php
```diff
@@ -49,29 +49,38 @@
...
                     }
                     if (count($error) == 0) {
-                        $success = $tteam->create($_GET["tournamentid"], $auth["userid"], $_POST['team_name'], $_POST["set_password"], $_POST['team_comment'], "team_banner");
+                        $success = $tteam->create($_GET["tournamentid"], $auth["userid"], $_POST['team_name'], $setPasswordParameter, $_POST['team_comment'], "team_banner");
                     }
                 }
 
                 if (count($error) == 0 and $success) {
                     // Update-League-IDs
-                    $tteam->UpdateLeagueIDs($auth["userid"], $_POST["wwclid"], $_POST["wwclclanid"], $_POST["nglid"], $_POST["nglclanid"], $_POST["lgzid"], $_POST["lgzclanid"]);
```
M4LuZ, line 74: I'm reasonably certain all of these leagues are dead and all the related code can be dropped

M4LuZ (approved):

modules/tournament2/join.php
andygrunwald: That is true. I do assume this would be a bigger change across the codebase. Hence, I would keep this excluded in this PR.
This is only "to get the current state somehow working properly".

PR #905, opened 2024-02-24
Github Actions: New workflow to lint Dockerfile

andygrunwald:
### What is this PR doing?

Introducing hadolint (https://github.com/hadolint/hadolint) to lint Dockerfiles.

This ensures we are applying best practices for our Docker setup.

None

### Checklist


Dockerfile
```diff
@@ -7,16 +7,16 @@ COPY . /code
...
-    && apt-get install -y \
-        libfreetype6-dev \
-        libjpeg62-turbo-dev \
-        libsnmp-dev \
-        snmp \
-        unzip \
-        libzip-dev \
-    && docker-php-ext-install -j$(nproc) mysqli snmp \
+    && apt-get install -y --no-install-recommends \
+        libfreetype6-dev=2.10.4+dfsg-1+deb11u1 \
```
M4LuZ, line 11: Do you think version pinning is a good idea in our usecase?
> Version pinning forces the build to retrieve a particular version regardless of what’s in the cache. This technique can also reduce failures due to unanticipated changes in required packages.

But If I'm not mistaken we do not have persistent cache due to the environment and we will most probably do not (want to) maintain minor library version upgrades and dependencies.

Yes - there is a risk of behavioural changes, but is it worth the effort?

andygrunwald: We don't have a docker build cache. This is executed at build stage of a docker image, not at run stage.

> Yes - there is a risk of behavioural changes, but is it worth the effort?

Good question. Personally, I am mostly into stability and unexpected behaviour. LANSuite doesn't go deep into the system, that's why I am also OK with ignoring https://github.com/hadolint/hadolint/wiki/DL3008

I am open for both. Let me know what you prefer.

M4LuZ: I'm also fine with either, but would personally prefer not to version pin:
* In my understanding only the latest minor update is kept in the source repos. Meaning that if there is an update from `deb11u1` to `deb11u2` and we've defined a fixed version then executions may regularly fail and we permanently have to bump minor version requirements. I understand that the change frequency in ubuntu is higher than debian, so this may happen quite a lot.
* Fixed versions have to be redefined again when we upgrade the OS release, but I think this is a case where it is a good idea to test and evaluate functionality as major version bumps may happen. But that is more a process requirement than a version pinning requirement

But we can also just start with it and see how it works. 
Removing the versions can be done at any time later if it turns out not to be a good idea.
Or maybe it just works (tm) ;)

I'll accept and leave it to you if you want to leave as-is or remove versions.

M4LuZ (approved):

M4LuZ:
Also: 
 CHANGELOG.md entry - Not needed
 Documentation update - Not needed

My over-careful self would say that a note should be added in both as there is a change in behaviour and people may need to be aware.

PR #908, opened 2024-02-24
Delete `ext_scripts/captcha.php` as it is not used

andygrunwald:
### What is this PR doing?

In https://github.com/lansuite/lansuite/pull/483, `ext_scripts/captcha.php` was touched.
I could not find any reference in the LanSuite code that is using this file.

Masterform is using `ext_scripts/ascii_captcha.class.php`.

Please double check before merging

None

### Checklist

- [X] `CHANGELOG.md` entry


M4LuZ:
As far as I can see this was either intended as replacement or alternative for the ascii captcha.
User registration form is the only place where I am aware of a captcha being used and that uses - as you mentioned - `ascii_captcha.class.php`

I got no clue if it is in any shape or form better than the text captcha and may be implemented instead of being deleted.
I guess there is a good reason why text captchas are not used anywhere else (and so we may also not), but I am lacking any facts to prove that.

andygrunwald:
@M4LuZ If we aim to replace the ASCII Captcha (which I would be in favor of), I would not replace it with `ext_scripts/captcha.php`. I would use one well maintained 3rd party lib from https://packagist.org/?query=captcha

There you have a more modern captcha and well maintained packages.
Hence, I am suggesting deleting this captcha implementation anyway.

What do you think?

Screenshot 2024-02-28 at 19 42 37

PR #912, opened 2024-02-25
[Discord] Rework Discord module to get it functional again

andygrunwald:
### What is this PR doing?

This Pull Request makes the Discord module working again + adds a few improvements:

* Set correct API Widget URL (the old one was not working anymore)
* Include `symfony/http-client` for proper HTTP Calls + error checking
* We delegate the error checking into the box part of the module
* We used dependency injection for caching and HTTP client to get the Discord class testable
* We only cache successful calls / No error messages are cached

### Screenshots

Error case
Screenshot 2024-02-25 at 12 34 16

Successful case
Screenshot 2024-02-25 at 13 10 40

None

### Checklist


M4LuZ (approved):
I do not like some of the design decisions, but basically I'm happy that the module is fixed and that trumps any other points ;)

But here's the feedback nevertheless:
* Initializing an instance of `httpClient` in `index.php` independent of the actual usage introduces a generic overhead on every single call even if we actually only use it (at the moment) in one place (and there also only if the cache is stale). 
There may be a few additional uses in the future, but I do not consider it a generic functionality needed often/everywhere. So maybe it would be better to instantiate late when needed?
* Getting this unit-testable is fine. Being able to do DI is also a fair point. Where I stumble a bit is that this appears to be designed for easy use for unit tests and not for easy use in development. I would see the priorities the other way around. 
Most of the calls will never need to pass a different/dedicated object. _Especially_ if we instantiate a global object to be used. And in that case we can also just instantiate the global object in the unit test. Why not make it an optional parameter?
 (But that is also me being burned by many lines of dependency-creating Java-code just to call a function in a throw-away object once)
* Same for the cache with the added point that  I am not sure of the behaviour of the cache object on pass-by-value calls. I guess it's fine, I haven't seen any warnings, but not sure.

Still happy to have this merged as-is to have the issues fixed independent on your views on my points ;)

andygrunwald:
Thanks @M4LuZ - Lets go through the feedback:

> Initializing an instance of httpClient in index.php independent of the actual usage introduces a generic overhead on every single call even if we actually only use it (at the moment) in one place (and there also only if the cache is stale).
There may be a few additional uses in the future, but I do not consider it a generic functionality needed often/everywhere. So maybe it would be better to instantiate late when needed?

I can create the instance in the Discord module for now.
I do expect more usage of this in the future, but we should not solve future problems.
About the "overhead": Because it is only an object instance, that is not holding any data, the proof in actual "overhead" usage of memory is questionable and still need to be proofed with data. I am strongly convinced that this is negligible.

> Getting this unit-testable is fine. Being able to do DI is also a fair point. Where I stumble a bit is that this appears to be designed for easy use for unit tests and not for easy use in development. I would see the priorities the other way around.
Most of the calls will never need to pass a different/dedicated object. Especially if we instantiate a global object to be used. And in that case we can also just instantiate the global object in the unit test. Why not make it an optional parameter?
(But that is also me being burned by many lines of dependency-creating Java-code just to call a function in a throw-away object once)

I strongly advise to move away from global variables or class objects created inside a function. If we aim for a more stable system, isolation is key, dependency injection is one way to solve it.
If we aim for "ease of use", we still can create a Factory to create all these objects and just call `Factory->create InstanceYXZ()`. Things like Symfony DependencyInjection Component is providing this functionality https://symfony.com/doc/current/components/dependency_injection.html

The class can't work without the HTTP lib, that's why this object is not optional. The caller (outside) don't know if the cache is stale or still fresh. And it should not know.
What we can do is to make the cache parameter optional. If the cache is not provided, caching is not provided, but the class is still usable and workable. Without HTTP lib, it is not functional.

> Same for the cache with the added point that I am not sure of the behaviour of the cache object on pass-by-value calls. I guess it's fine, I haven't seen any warnings, but not sure.

Would you mind elaborating on this? I am not sure if I can follow you.

andygrunwald:
@M4LuZ Let me know what you think - I won't merge it before.

andygrunwald:
@M4LuZ An chance that you have a look at this one?

andygrunwald:
I had a quick chat with @M4LuZ - He is fine with merging this.
A general design discussion will be started by him.

PR #914, opened 2024-02-27
Fixing several PHP errors in modules `stats`, `seating`, `tournament` and `party`

andygrunwald:
### What is this PR doing?

Fixing several PHP errors in modules `stats`, `seating`, `tournament` and `party`

None

### Checklist

### DIFF
--- a/modules/party/plugins/usrmgr_details_main.php
+++ b/modules/party/plugins/usrmgr_details_main.php
@@ -3,20 +3,23 @@
 // This File is a Part of the LS-Pluginsystem. It will be included in
 // modules/usrmgr/details.php to generate Modulspezific Mainpage-entrys
 // (fielsets) for Userdetails
-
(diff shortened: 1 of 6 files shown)


modules/party/plugins/usrmgr_details_main.php
```diff
@@ -6,20 +6,25 @@
 
 $user_party = $db->qry_first('
```
M4LuZ, line 7: It would be good to implement pending refactoring changes (as here switching from $db to $database) whenever we touch code. A test for function has to be done anyways

andygrunwald: Fixed in https://github.com/lansuite/lansuite/pull/914/commits/519b4f653232c790ea0b3f3e215173055f7b4c70

M4LuZ (approved):

PR #923, opened 2024-03-01
[usrmgr] Add missing empty and error check

rstephan:
### What is this PR doing?

If you call `http://.../index.php?mod=usrmgr&action=pwrecover&step=2`  out of nowhere you get a warning about an undefined array key.
```
PHP Warning: Undefined array key "pwr_mail" in /www/htdocs/xxx/modules/usrmgr/remind.php on line 11
PHP Warning: Trying to access array offset on value of type bool in /www/htdocs/xxx/modules/usrmgr/remind.php on line 12
PHP Warning: Trying to access array offset on value of type bool in /www/htdocs/xxx/modules/usrmgr/remind.php on line 14
```

### DIFF
--- a/modules/usrmgr/remind.php
+++ b/modules/usrmgr/remind.php
@@ -8,10 +8,13 @@ if (!$cfg['sys_internet']) {
     switch ($stepParameter) {
         // Email prüfen, Freischaltecode generieren, Email senden
         case 2:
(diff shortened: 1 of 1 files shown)


andygrunwald (approved):
Thanks!

PR #926, opened 2024-03-02
924-fix/improve cronjob handling

### LINKED ISSUE #924: Unexecutable cron jobs lead to blank page in some cases
## Expected Behavior

Cron jobs should run in the background and not affect user interaction.
Issues with cronjobs should be clearly visible for admins and manageable via the related functionality

## Current Behavior

Inexecutable sql statements lead to blank pages, as errors are thrown e.g. in the user registration form
Additionally as jobs are also run on the cron config dialogue, preventing administrators from editing them to fix a potential issue.
outcome of last execution is not logged

## Possible Solution

* Add try-catch block for jobs
* add field for last error output
* prevent cron from running on cron-config pages
* _(optional) stop execution on multiple failed executions_
* _(optional) fix field `active` not actually being used_

* try to register on page with incorrect sql in job
* try to edit failing job on cron2 admin page

## Context

in this case there was a job that did not have the correct ls prefix defined and thus lead to an SQL error on execution.
This prevented an user from registration

## Your Environment

* Version used: git/master
* Operating System and version: debian 11/PHP 8.1 / MySQL 8.0
* Enabled features: cron2

M4LuZ:
### What is this PR doing?

This PR introduces multiple fixes / changes:
* Cronjobs are only executed when marked as active
* new fields introduced for `last_error`, `error_runs` to show if action failed and what the error message was 
* Added field `runtime` to show total script run time on successful execution
* generic errors thrown by the cron script are catched not to influence further script execution
* Failed jobs are deactivated after three consecutive failed executions
* Refactored to use `$database` instead of `$db`

Fixes #924

### Checklist

- [x] `CHANGELOG.md` entry
- [x] ~Documentation update~


modules/cron2/mod_settings/db.xml
```diff
@@ -59,6 +59,14 @@
 				<key></key>
 				<default></default>
 				<extra></extra>
+			</field>
+				<field>
```
andygrunwald, line 63:
```suggestion
			</field>
			<field>
```

modules/cron2/Classes/Cron2.php
```diff
@@ -4,48 +4,109 @@
...
         }
     }
+
+    /**
+     * Deactivates a job with the given ID
+     *
+     * @param int $jobid ID of the job to be deactivated
+     * @return bool true if job was found, false if not
+     */
+    public function deactivateJob(int $jobid = 0)
```
andygrunwald, line 105:
```suggestion
    public function deactivateJob(int $jobid = 0): bool
```

Lets use the type system of the language

modules/cron2/Classes/Cron2.php
```diff
@@ -4,48 +4,109 @@
 
 class Cron2
 {
+
+    const LS_CRON_MAX_ERRORS = 3; //Maximum amount of errors for a job
```
andygrunwald, line 8: We don't need to prefix the constant with `LS_`, because this constant only exists in the scope of the `Cron2` class.

I suggest to rename this constant to 
```suggestion
    /**
     * Maximum amount of errors for a job
     */
    const CRON_MAX_ERRORS = 3;
```
and using PHP doc

modules/cron2/Classes/Cron2.php
```diff
@@ -4,48 +4,109 @@
...
+        );
+
         if ($row != false) {
-            if ($row['type'] == 'sql') {
-                $sql = str_replace('%prefix%', $config['database']['prefix'], $row['function']);
-                $db->qry('%plain%', $func->AllowHTML($sql));
-            } elseif ($row['type'] == "php") {
-                require_once 'ext_scripts/'.$row['function'];
+            try {
+                $runtime = microtime(true);
```
andygrunwald, line 31: One of the false-hood believes of programmers is that time always runs forward. For time measurement, we should use monotonic clocks (via https://www.php.net/manual/en/function.hrtime.php) and not the UNIX timestamp. Time can go backwards - BUt we want to measure the duration.
See https://blog.codeminer42.com/the-monotonic-clock-and-why-you-should-care-about-it/

modules/cron2/Classes/Cron2.php
```diff
@@ -4,48 +4,109 @@
...
+                    SET
+                        lastrun = NOW(),
+                        error_runs = error_runs + 1,
+                        last_error = ?
+                    WHERE jobid = ?',
+                    [$e->getMessage(), $jobid]
+                );
+                $func->log_event(t('Die Ausführung von Job %1 ist fehlgeschlagen', $row['name']), 2);
+
+                if ($row['error_runs'] >= $this::LS_CRON_MAX_ERRORS) {
```
andygrunwald, line 66:
```suggestion
                if ($row['error_runs'] >= self::LS_CRON_MAX_ERRORS) {
```

andygrunwald (changes requested):
Looks pretty good. I suggested a few changes.

modules/cron2/Classes/Cron2.php
M4LuZ: that was more a case of "let's not forget again to do that" ;)
Thanks for catching that!

modules/cron2/Classes/Cron2.php
M4LuZ: This was less an assumption of getting a good and exact measurement, but more a _good enough_ time for rough display purposes.
So while it is fit for purpose as-is it is not a big deal to rework to use `hrtime`

andygrunwald (approved):

PR #930, opened 2024-03-03
[Usrmgr] Remove ICQ field

andygrunwald:
### What is this PR doing?

I consider the ICQ messenger as dead.
Lets remove this field all over from the codebase.

None

### Checklist

- [X] `CHANGELOG.md` entry


M4LuZ (approved):
> I consider the ICQ messenger as dead.

BLASPHEMY!
95044945!

I would say MSN would be even more worth deleting, but I guess that is a follow-up PR then? ;)

andygrunwald:
> I would say MSN would be even more worth deleting, but I guess that is a follow-up PR then? ;)

Yes, one change after another :) 
I will merge this and continue with removing MSN as well.

PR #941, opened 2024-03-10
[XSS] Prevent XSS possibility in http-equiv="refresh" header

andygrunwald:
### What is this PR doing?

[XSS] Prevent XSS possibility in http-equiv="refresh" header

None

### Checklist

- [X] `CHANGELOG.md` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,7 @@ We could not reconstruct _all_ changes, but we tried our best to make the most o
 
 - [Database] Introduced a new Database class with first-class prepared statement support
 - [System] Protect module inclusion for path traversal
+- [XSS] Prevent XSS possibility in http-equiv="refresh" header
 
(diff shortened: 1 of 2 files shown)


M4LuZ (approved):
Good to have this fixed.
Based on my understanding this requires a prior compromise to be usable and then does not provide anything that cannot be done with the original injection.

Thus I do not see the need for a seprate security bulletin.
I hope you agree?

andygrunwald:
> Good to have this fixed. Based on my understanding this requires a prior compromise to be usable and then does not provide anything that cannot be done with the original injection.
> 
> Thus I do not see the need for a seprate security bulletin. I hope you agree?

Yes and no. It is a typical XSS vuln.
Sending a link to you is enough, but the vuln is not persistent.
It is only "client" side.

Like `?sitereload="><script>some bad JS</script>` and then you need to click on it.

Issue #409, opened 2019-01-27
credits.php needs rework / updates


aj-gh:
[credits.php](https://github.com/lansuite/lansuite/blob/master/modules/about/credits.php) should be updated as it contains several outdated/broken links in prominent places.
The listed "tested" PHP and MySQL versions are also quite outdated.
Also the hardcoded values for "lines of code" don't really make sense anymore I guess.

Similar to #398.

## Expected Behavior

Updated information ;)

## Current Behavior

Stone-age details.

PR #942, opened 2024-03-10
[About] Removed PHP and HTML Line count from about site

andygrunwald:
### What is this PR doing?

[About] Removed PHP and HTML Line count from about site

I decided to remove this functionality because
* it is out of date
* it serves not a single purpose

Related #409

### Checklist

- [X] `CHANGELOG.md` entry


andygrunwald:
@M4LuZ Conflict resolved

andygrunwald:
@M4LuZ Conflict resolved

M4LuZ:
Your argumentation-foo is weak. ;)

> it is out of date

given that it provides the means to update the count this is only true if not executed at any time.

> it serves not a single purpose

It serves the purpose of showing LOC. That's what it is for, that is what it does.
And that has the underlying purpose of showing the sheer size of the project.

What I do agree with:
* Having this done not as part of the install / build process or as an automated job is not good, that leads to the observed risk of having outdated information.
* This is completely nonessential

Thus even though I cannot see an argument to remove it, I also have no strong reason to oppose it.
If it helps you then I'm fine to remove it

M4LuZ (approved):

andygrunwald:
> It serves the purpose of showing LOC. That's what it is for, that is what it does.
> And that has the underlying purpose of showing the sheer size of the project.

Think about the "customer" expectation: The "customer" is seeking a Lan Party Content management system. Why is the size of the project relevant for this need?

This system is complex and has too much functionality that is considered as nonessential. We can run a "Jack of all trades" system or a specialized system that does a perfect job. A jack of all trade system with the current time + engineering power is leading to a system that tries a lot but does nothing good. I feel we are in this situation and need to do a compromise.

M4LuZ:
That's a fair argument for removal that I am happy to agree with. 
Thanks for clarifying that.

PR #945, opened 2024-03-10
[About] Remove list of Thank You to various open source projects

andygrunwald:
### What is this PR doing?

This PR removes the list of (random) thanks to various open source projects.
I like the idea, however, we, lansuite, are also open source and we build upon others.

I think we should remove this, because
* we are never able to keep this list up to date
* where do we stop? Should we thank github for providing the platform? our parents to gave us life?
* it doesn't serve any purpose
* every dev is using a different tooling

Related #409

### Checklist

### DIFF
--- a/modules/about/credits.php
+++ b/modules/about/credits.php
@@ -28,33 +28,6 @@ $dsp->AddSingleRow('Wir danken allen Modulprogramierern.</b><br /> Im Modulkonfi
 $dsp->AddSingleRow('<b>On-line Documentation</b>', 'align="center"');
 $dsp->AddSingleRow('Sebastian Haslinger', 'align="center"');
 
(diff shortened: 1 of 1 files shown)


andygrunwald:
@M4LuZ Conflict resolved.

M4LuZ:
Hmm. I see that page as a piece of the by now 17? year long history of LanSuite and yes, while it is not up to date and the list of thanks is quite a bit random, it is kind of nice to see what the spirit was _back then_

It does serve a purpose: It shows grattitude from pior developers in regards to what helped them with their work and it shows our gratitude to preserve that.
Nevertheless it does make sense to have a page that is current

Maybe throw it as legacy documentation onto the website?

andygrunwald:
@M4LuZ Any possibility for a review/decision on this?

PR #946, opened 2024-03-10
[About] Move contributors to CONTRIBUTORS.md and add current people to the Thank You list

andygrunwald:
### What is this PR doing?

[About] Move contributors to CONTRIBUTORS.md and add current people to the Thank You list

Related #409

### Checklist

- [X] `CHANGELOG.md` entry
- [X] Documentation update


CONTRIBUTORS.md
```diff
@@ -0,0 +1,55 @@
+# Thanks for your effort
+
+Thanks goes to all contributors to this project.
+
+Have a look at https://github.com/lansuite/lansuite/graphs/contributors to see all contributors.
```
M4LuZ, line 5:
```suggestion
Have a look at https://github.com/lansuite/lansuite/graphs/contributors to see current contributors.
```
As listing does not show unmapped SVN accounts

modules/about/credits.php
```diff
@@ -4,26 +4,8 @@
...
-$dsp->AddSingleRow('<b>Erweiterungen, Optimierungen, Fehlerbeseitigung</b>', 'align="center"');
-$dsp->AddSingleRow('Raphael Mansmann<br />Jochen Jung<br />Marco Müller', 'align="center"');
-
-$dsp->AddSingleRow('<b>Modul-Programming</b>', 'align="center"');
-$dsp->AddSingleRow('Wir danken allen Modulprogramierern.</b><br /> Im Modulkonfigurator kann man nachsehen wer welches Modul geschrieben hat', 'align="center"');
-
-$dsp->AddSingleRow('<b>On-line Documentation</b>', 'align="center"');
-$dsp->AddSingleRow('Sebastian Haslinger', 'align="center"');
+$dsp->AddSingleRow('<b>Thanks to all project contributos</b>', 'align="center"');
+$dsp->AddSingleRow('Please head over to <a href="https://github.com/lansuite/lansuite/blob/master/CONTRIBUTORS.md" title="All lansuite contributors">project contributors</a> to see the full list of project contributors.', 'align="center"');
```
M4LuZ, line 8: What do you think about including the page in the webpages generated by docusaurus?
Also same as above: Git AFAIK lists current contributors, not a full list of all people ever contributing

M4LuZ (approved):
I'm generally fine with this, see two comments before merging.

Scratching my head a bit why this was considered to be necessary now, though ;)

andygrunwald:
> Scratching my head a bit why this was considered to be necessary now, though ;)

It was not fully necessary, still improves the product in a little bit.

modules/about/credits.php
andygrunwald: Not all contributors are listed, this is true. I think we will never find all contributors again. What we have now is all what we got so far.

Not sure if we need to add this to the documentation. I think mostly that this is information from/for development.

PR #955, opened 2024-03-15
Fixed broken end-date check

rstephan:
### What is this PR doing?

The End-Date should be "bigger" than the Start-Date, right?

Fixes broken end-date check

### Checklist

- [ ] `CHANGELOG.md` entry
- [ ] Documentation update

### DIFF
--- a/modules/party/Functions/CheckEndDate.php
+++ b/modules/party/Functions/CheckEndDate.php
@@ -7,7 +7,7 @@ function CheckEndDate($enddate): bool|string
 {
     global $func;
 
-    if ($func->str2time($enddate) < $func->str2time($_POST['startdate'])) {
+    if ($func->str2time($enddate) > $func->str2time($_POST['startdate'])) {
         return t('Der Endzeitpunkt muss nach dem Startzeitpunkt liegen');
     } else {
         return false;


modules/party/Functions/CheckEndDate.php
```diff
@@ -7,7 +7,7 @@ function CheckEndDate($enddate): bool|string
 {
     global $func;
 
-    if ($func->str2time($enddate) < $func->str2time($_POST['startdate'])) {
+    if ($func->str2time($enddate) > $func->str2time($_POST['startdate'])) {
         return t('Der Endzeitpunkt muss nach dem Startzeitpunkt liegen');
```
andygrunwald, line 11: The message `Der Endzeitpunkt muss nach dem Startzeitpunkt liegen` indicates an error, that `$enddate` is earlier than start date.
If we switch it, isn't this the wrong way?

rstephan:
The function `$func->str2time()` is the problem.
If you put a single-digit time into it, the result is partially filled with the current time.
The string  `2024-04-01 0:0:00` will be 1711983360 -> `2024-04-01 14:56:00` at 15:56:21 (timezone is also not handled properly) at that time.
If you call the function a minute later the result will be different!

Therefore this pull request is not sufficient to fix the problem, and can be closed.

andygrunwald:
@rstephan I will move on and close this PR. Thanks!

PS: I created https://github.com/lansuite/lansuite/issues/971 to fix the problem with `$func->str2time()`

PR #960, opened 2024-03-15
[Tournament2] Switch many queries to Prepared Statements API

andygrunwald:
### What is this PR doing?

[Tournament2] Switch many queries to Prepared Statements API

This is not covering all statements.
But it is a first step.

None

### Checklist


M4LuZ (approved):
> But it is a first step.

Given the amount of changes that is rather a small jump, many thanks for taking the time for that.

PR #961, opened 2024-03-15
[Func] Remove `escape_sql`

andygrunwald:
### What is this PR doing?

[Func] Remove `escape_sql`

Addslashes is not a solid method to avoid SQL injection.

None

### Checklist


andygrunwald:
@M4LuZ conflict resolved.

M4LuZ (approved):
As per discussion: There was a question if there was any slash-removal at some db-reading function which could have caused formatting issues. But it appears that content had slashes added before being written but never having it removed

PR #964, opened 2024-03-16
[Auth] Switch Auth class to prepared statements

andygrunwald:
### What is this PR doing?

[Auth] Switch Auth class to prepared statements

None

### Checklist

- [X] `CHANGELOG.md` entry
- [X] Documentation update


inc/Classes/Auth.php
```diff
@@ -86,14 +86,18 @@ class Auth
...
+     * Database
+     */
+    private Database $database;
+
     /**
      * auth constructor.
      * @param string $frmwrkmode Frameworkmode for switch Stats
      */
-    public function __construct(Request $request, $frmwrkmode = '')
+    public function __construct(Database $database, Request $request, $frmwrkmode = '')
```
M4LuZ, line 98: Hmm. 
Do you see any chance that we are using a different DB instance or solution in the future?
Otherwise the requirement to provide this without standard value does not make sense.
Also - if there is a good case for this - would it be better to move it at the end and make it optional so that existing calls work without modification?

M4LuZ (approved):

inc/Classes/Auth.php
andygrunwald: > Do you see any chance that we are using a different DB instance or solution in the future?

No, not in a production system.
However, this is a pattern called [Dependency Injection](https://en.wikipedia.org/wiki/Dependency_injection) and it allows us to start writing unit tests for Auth class without having a database, because we can inject a object that acts like a database.

The alternative would be the database as global variable, which is also a very bad practice. LanSuite suffers from global variables and side effects already. See https://www.baeldung.com/cs/global-variables for some basic reasons.

On top of this: This marks the dependency as explicit and it is a preferred way + good practice in software engineering.

I think it is good for now. If we want to change this, we can do it in another pull request.

M4LuZ: I'm well aware that it is DI and and I'm also well aware that global variables are not without their issues and thus should be used judiciously.
Both have advantages and disadvantages, and the choice of implementation should done based on how well they fit the requirements.
My question was in regards to the specific choice for implementation, as implementation causes disadvantages (from first glance also without need) without needing or providing the benefits that are the key strengths of DI (injecting a mock DB object requires same amount and complexity of code in both cases). 
Also given that it is not _just_ the choice between pure global variables and DI, but any variant in between and on the side. 

But at the end of the day I already accepted given that there are just two uses and the other changes are far too beneficial to keep this open due to that.

PR #965, opened 2024-03-16
[PDF] Rework of PDF module with focus on Certificates

andygrunwald:
### What is this PR doing?

[PDF] Rework of PDF module with focus on Certificates

None

### Checklist

### DIFF
--- a/modules/pdf/Classes/PDF.php
+++ b/modules/pdf/Classes/PDF.php
@@ -137,6 +137,7 @@ class PDF
         $this->data_type_array['guestcards']['user_block']      = "Sitzblock";
         $this->data_type_array['guestcards']['user_ip']         = "IP-Adresse";
         $this->data_type_array['guestcards']['party_name']      = "Lanparty-Name";
+        $this->data_type_array['guestcards']['party_date']      = "Lanparty-Datum";
         $this->data_type_array['guestcards']['plz']             = "PLZ";
         $this->data_type_array['guestcards']['city']            = "Ort";
         $this->data_type_array['guestcards']['birthday']        = "Geburtstag";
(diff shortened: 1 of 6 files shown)


M4LuZ (approved):
I have absolutely zero clue how the PDF module works, thus unable to review the datails and trusting them to be OK. superficially this looks fine.

Given that apparently document templates are changed a remark in both changelog and update documentation may be warranted?

Speaking of documentation: I guess this is a module that would warrant to see if we can transfer some of the old documentation to docusaurus

andygrunwald:
> Given that apparently document templates are changed a remark in both changelog and update documentation may be warranted?

This is a good point.
I am digging into the template parts again and see if this is needed.

> Speaking of documentation: I guess this is a module that would warrant to see if we can transfer some of the old documentation to docusaurus

Good point. Where can we find the "old" documentation?

M4LuZ:
> Good point. Where can we find the "old" documentation?

Internet Archive for lansuite.orgapage.de ist the best worst solution till we work on getting the old page back up or content imported otherwise

M4LuZ:
Hi @andygrunwald, just to make sure that you are aware: This has been approved already but I did not merge in case entries for changelog and documentation are needed

andygrunwald:
@M4LuZ Thanks - I will take care of this PR. We updated the docs inside LanSuite for the PDF module which is still to be added to the PR. A few changes are coming.

andygrunwald:
@M4LuZ I am added the latest changes + marked those in the changelog.

Most of these changes are related to certificates to make them really work.
Along those things, small items like in module documentation have been added, see

Screenshot 2024-04-11 at 09 56 45

As aligned with you, I am merging those.

PR #991, opened 2024-03-26
[Sponsor] Switch SQL statements to Prepared Statements API

andygrunwald:
### What is this PR doing?

[Sponsor] Switch SQL statements to Prepared Statements API

Related #183

### Checklist


modules/sponsor/show.php
```diff
@@ -7,7 +7,7 @@
 $dsp->NewContent(t('Unsere Sponsoren'), t('Bei den folgenden Sponsoren möchten wir uns herzlich für ihren Beitrag zu unserer Veranstaltung bedanken.'));
 $sponsoren = $db->qry("SELECT * FROM %prefix%sponsor WHERE sponsor ORDER BY pos, sponsorid");
 
-$db->qry("UPDATE %prefix%sponsor SET views = views + 1 WHERE sponsor");
+$database->query("UPDATE %prefix%sponsor SET views = views + 1 WHERE sponsor");
```
M4LuZ, line 10: I'm surprised that this is OK in regards to SQL, the `WHERE sponsor` clause is either nonsensical or unnecessary. 
From how I understand the logic the display value for all sponsors should be increased by 1 as the sponsors page shows all of them.

andygrunwald: `sponsor` is an actual database field:
```
<field>
  <name>sponsor</name>
  <type>int(1)</type>
  <null></null>
  <key></key>
  <default>1</default>
  <extra></extra>
</field>
```

The way I understand this: It indicates if a sponsor is active or not active. Hence the `WHERE sponsor` is a boolean condition.

In https://github.com/lansuite/lansuite/pull/991/commits/f12d0cb1f8795b6198185405997c85953606e512 I updated to make it more clear. From my check, there is no functionality to "deactivate" a sponsor right now, but I would not remove the field - Rather check if we want to do it later and improve on this.

M4LuZ: That's fine and thanks for the explanation

M4LuZ (approved):

PR #1006, opened 2024-03-26
[Install] Switch SQL statements to Prepared Statements API

andygrunwald:
### What is this PR doing?

[Install] Switch SQL statements to Prepared Statements API

Related #183

### Checklist


M4LuZ (approved):

M4LuZ (approved):

modules/install/Classes/Import.php
```diff
@@ -666,13 +659,13 @@ public function ImportLanSuite($del_db, $replace, $no_seat, $signon, $comment)
...
                         // Search clan
-                        $search_clan = $db->qry_first("SELECT clanid FROM %prefix%clan WHERE name = %string%", $clan);
+                        $search_clan = $database->queryWithOnlyFirstRow("SELECT clanid FROM %prefix%clan WHERE name = ?", [$clan]);
                         if ($search_clan['clanid'] != '') {
                             $clan_id = $search_clan['clanid'];
 
                         // Insert new clan
                         } else {
-                            $db->qry("INSERT INTO %prefix%clan SET name = %string%, url = %string% ", $clan, $clanurl);
+                            $database->query("INSERT INTO %prefix%clan SET name = ?, url = ?", [$clan, $clanurl]);
```
M4LuZ, line 668:
```suggestion
                            $database->query("INSERT INTO %prefix%clan SET name = ?, url = ?", [$clan, '']);
```
`$clanurl` appears uninitialised in this scope.
(It makes sense from the context,  just users are imported, but not clans created during the event, thus there is just the name that can be initialized

PR #1024, opened 2024-03-26
[Functions] Switch SQL statements to Prepared Statements API

andygrunwald:
### What is this PR doing?

[Functions] Switch SQL statements to Prepared Statements API

Related #183

### Checklist


inc/Functions/CheckValidEmail.php
```diff
@@ -11,7 +11,7 @@
  */
 function CheckValidEmail($email)
 {
-    global $cfg,$db;
+    global $cfg, $db, $database;
```
M4LuZ, line 14: Function uses neither `$db nor `$database`
```suggestion
    global $cfg
```

andygrunwald:
@M4LuZ PR adjusted. Implemented a connection status into the database class to use it in the Error handler.

inc/Classes/Database.php
```diff
@@ -72,10 +77,27 @@ public function connect(): bool
...
         $this->setCharset($this->charset);
+        $this->setConnectionStatus(true);
 
         return true;
     }
 
+    /**
+     * Sets the connection status
+     */
+    private function setConnectionStatus(bool $status): void
```
M4LuZ, line 88: I don't see the added value of a private setter function especially given the simple code it contains. 
But fine for me as code is working and still better than what we had ;)

M4LuZ (approved):

PR #963, opened 2024-03-16
[Framework] Rework of the Framework class

andygrunwald:
### What is this PR doing?

This Pull Request reworks the Framework class.

It mainly does:
- Unified and more explicit renaming of the methods (e.g. JS vs. JavaScript)
- Usage of dependency injection (e.g. for the template engine and the debug class)
- Usage of Symfony HTTP Foundation and not requiring $_SERVER anymore
- Ensuring all required template variables are set
- Usage of class constants for standard values

I suggest to review commit by commit

### Which issue(s) this PR fixes:

None

### Checklist

- [X] `CHANGELOG.md` entry - Not needed
- [X] Documentation update - Not needed

PR #1035, opened 2024-03-26
[Framework] Switch SQL statements to Prepared Statements API

andygrunwald:
### What is this PR doing?

[Framework] Switch SQL statements to Prepared Statements API

Related #183

### Checklist


M4LuZ (changes requested):
There is no actual switch happening here.
And the only call to `$db` is in Line 379 counting the amount of queries for page generation where we have no equivalent in `$database`.
So we either need to create that or remove line and both global variable references

andygrunwald:
@M4LuZ You are right - I am fine with leaving this as a draft, because it kind of conflicts with https://github.com/lansuite/lansuite/pull/963

Hence, I suggest to first merge https://github.com/lansuite/lansuite/pull/963 and than I re-take this one.

andygrunwald:
Closing this one due to not an actionable change.

PR #1036, opened 2024-03-26
[Templating] Upgrading from v4 to v5 (Smarty)

andygrunwald:
### What is this PR doing?

Applying the upgrade guide: https://smarty-php.github.io/smarty/stable/upgrading/

We do not change the caching strategy on purpose, because this needs further evaluation as the doc says it can be tricky.

None

### Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "ext-zlib": "*",
         "ext-ftp": "*",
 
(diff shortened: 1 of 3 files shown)


andygrunwald:
I merged this with force, because due to https://github.com/lansuite/lansuite/pull/987 the master was broken and the index.php threw a PHP Fatal Error.

PR #1056, opened 2024-03-31
[MasterSearch2] Fix wrong number of rows found

andygrunwald:
### What is this PR doing?

[MasterSearch2] Fix wrong number of rows found

The issue: `FOUND_ROWS` refers to the last SQL statement, which is made with the old DB class. This query can't be easily switched over to the new DB class, due the current usage.

None

### Checklist

### DIFF
--- a/modules/mastersearch2/Classes/MasterSearch2.php
+++ b/modules/mastersearch2/Classes/MasterSearch2.php
@@ -559,7 +559,7 @@ class MasterSearch2
 
         // Generate Page-Links
         $count_pages = 0;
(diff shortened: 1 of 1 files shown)


M4LuZ (approved):
Well...it should be easily switchable by changing the query above to also use `$database`.
The issue is almost certainly caused by `$db` and `$database` using two separate DB connections and thus having separate last queries and thus related counts.

We could think about having them share one connection, which would at least these issues in regards to partial changeover but that can cause other problems are both are not designed for shared use.

But given that this currently breaks functionality and no benefit is provided by using `$database` unless underlying query is also restructured I would merge and then have another go later.

PR #1069, opened 2024-04-01
[Seating] Fix undefined variable `$new_user`

andygrunwald:
### What is this PR doing?

[Seating] Fix undefined variable `$new_user`

None

### Checklist


modules/seating/seatadmin.php
```diff
@@ -9,10 +9,28 @@
 if ($stepParameter > 1 && (!$_GET['userid'])) {
     $func->error(t('Es wurde kein Benutzer ausgewählt'), "index.php?mod=seating&action=seatadmin");
 }
+
 if ($stepParameter > 2 && (!$_GET['blockid'])) {
     $func->error(t('Es wurde kein Sitzblock ausgewählt'), "index.php?mod=seating&action=seatadmin&step=2&userid={$_GET['userid']}");
 }
 
+
+$userIDParameter = $_GET['userid'] ?? 0;
```
M4LuZ, line 18: Hi @andygrunwald, this was needed to solve a warning for what case?
I fear that this may cause the following code to execute OK for userid = 0 in e.g. step10 (assign seat to user) when no userid is provided (intentional or not).

andygrunwald: Good catch. I moved this up in https://github.com/lansuite/lansuite/pull/1069/commits/d4c50382b2775b466f1a54d06f9b08b4bb778d2d

In this check: https://github.com/lansuite/lansuite/blob/1cf60caa87f3541ea872b85fafd3145fbf25d73e/modules/seating/seatadmin.php#L9-L11

It is now caught if the user id is 0 and the step = 10

PR #1080, opened 2024-04-05
[Shoutbox] Fix PHP warnings

andygrunwald:
### What is this PR doing?

[Shoutbox] Fix PHP warnings

Somehow, those got lost in the codebase while merging + resolving conflicts.
These are the lines previously introduced: https://github.com/lansuite/lansuite/commit/3154cda6b1856a683e9efd534c0b6870e55c4acc#diff-428113b8c26a45711199b45cb58e726707b5682d8b539172c2ab4e32592cea9b

None

### Checklist

### DIFF
--- a/modules/shoutbox/boxes/shoutbox.php
+++ b/modules/shoutbox/boxes/shoutbox.php
@@ -9,6 +9,10 @@ if (!$auth['userid']) {
     $userid = $auth['userid'];
 }
 
+$shoutDelay = $cfg['shout_delay'] ?? 20;
+$shoutEntries = $cfg['shout_entries'] ?? 5;
+$shoutLength = $cfg['shout_length'] ?? 160;
+
 $framework->addJavaScriptCode('var shoutdelay = ' . intval($shoutDelay) . ';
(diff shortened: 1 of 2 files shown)


M4LuZ (approved):
PR is fine, but now looking at this the second time the original issue of missing variable content was most probably caused by a missing DB update.

`config.xml` does define standard values, so by default these should always have a value unless actively removed.
Not sure if it a good choice to have default values in two locations as it may lead to change anomalies later on.

What do you think? 
Would it make sense to change config loading to set standard values if values are left undefined? 
Or do that on access by changing `$cfg` to a class object with related set/get functions that take care of that implicitly?

andygrunwald:
> Would it make sense to change config loading to set standard values if values are left undefined?

Yes, 100%.

> Or do that on access by changing $cfg to a class object with related set/get functions that take care of that implicitly?

This would also be my approach. One Configuration class to contain all types of configurations inside LanSuite. I don't have time to do this right now, but I keep it in my head.

M4LuZ:
Improvement raised as #1082

PR #1098, opened 2024-04-10
move ext-ftp and ext-snmp to suggested packages

M4LuZ:
### What is this PR doing?

Both the FTP and  SNMP modules are currently marked as hard dependencies in composer, although both are rather optional:
`SNMP` is not used unless the NOC module is active
`FTP` is not used unless either the download module is used with an external file location or the server module with a server configured as FTP

This moves both dependencies to the recommendations in `composer.json` 

None, issue was reported via Discord

### Checklist

- [x] ~`CHANGELOG.md` entry~  (I would say it is not necessary, as we introduced this dependency with a change before)
- [x] Documentation update - Corrected `requirements.md` with actual dependencies

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,10 +15,8 @@
         "php": "^8.1",
         "ext-gd": "*",
         "ext-mysqli": "*",
-        "ext-snmp": "*",
         "ext-json": "*",
         "ext-zlib": "*",
-        "ext-ftp": "*",
 
         "smarty/smarty": "^5.0",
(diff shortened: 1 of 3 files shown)


M4LuZ:
Need to revert `composer.lock` as update was done with php 8.2, which updated dependencies to incompatible versions.
I'll learn....someday....

andygrunwald (approved):
I need to ran `composer update ext-snmp` to update composer.lock - Otherwise good.

PR #1105, opened 2024-04-11
Update documentation to remove NO_AUTO_CREATE_USER

### LINKED ISSUE #1092: Upgrade from LanSuite v4.2 to v5.0-dev
It´s mentioned [there ](https://lansuite.github.io/lansuite/docs/guides/upgrade/)to put in the following line to /inc.base.config.php
``` 
[...]
[database]
sqlmode = "NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION"
[...]
```
Afterwards I get the following error on the Lansuite site
```
Die Verbindung zur Datenbank ist fehlgeschlagen. LanSuite wird abgebrochen. Zurückgegebener MySQL-Fehler: Variable 'sql_mode' can't be set to the value of 'NO_AUTO_CREATE_USER' (1231)
```

M4LuZ:
### What is this PR doing?

Updates update documentation as update with updated MySQL updated SQL modes.

Or in other words: 
`NO_AUTO_CREATE_USER` was removed with [MySQL 8.0](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-11.html#mysqld-8-0-11-deprecation-removal), thus documentation needs to be corrected.
Also added section to run through setup again, as takes care of a lot of steps otherwise needed (including writing the SQL mode as we need)

Fixes #1092 

### Checklist

- [ ] ~`CHANGELOG.md` entry~
- [x] Documentation update

### DIFF
--- a/website/docs/guides/guides-upgrade.md
+++ b/website/docs/guides/guides-upgrade.md
@@ -6,7 +6,7 @@ sidebar_position: 1
 
 ## Prerequisites
 
-* Your target system need to run PHP >= 8.1 and MySQL >= 5.7
+* Your target system needs to run PHP >= 8.1 and MySQL >= 5.7
 * A fair bit of time
 
 ## Preparation
@@ -72,6 +72,18 @@ Like:
 
 Copy them to the same place on the new installation, overwriting everything there.
 
+### Configuration: `environment.configured`
+
+In order to simplify some configuration steps and updates it is recommended to reset the configuration back to unconfigured and run through the installer again.
+This updates tables and menu entries and configures some standard options.
+This can be done by editing the following line in `inc/base/config.php`
+```
+[...]
+[environment]
+configured = 1
+[...]
+```
+...just *DO NOT select the option to overwrite the existing DB*
 ### Run specific release tasks
 
 It may be required to run additional steps to prepare everything for the new version.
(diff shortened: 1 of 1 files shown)


andygrunwald:
That is looking good.
What do you think about adding this setting also into the Install wizard as not accepted SQL mode?

https://github.com/lansuite/lansuite/blob/fefbaef096191f4b34297fa8fc8b1a3992308644/modules/install/Classes/Install.php#L677-L688

M4LuZ:
> What do you think about adding this setting also into the Install wizard as not accepted SQL mode?

`NO_AUTO_CREATE_USER` deactivates that a new user account is automatically created when granting a non-existent database user some access rights.

So it does not matter to us if it is present or not and thus does not need to be added to the disable-list. 
Additional reasoning: It is gone in more recent versions anyways and more of a do-set than don't-set option (as it prevents random user creation by spelling mistakes).

But it would make sense to add a note that option should be removed if already set in configuration file

andygrunwald (approved):

andygrunwald:
Thanks, this make sense.

Issue #1094, opened 2024-04-10
Profile picture is gone when editing the profile


Pimmal:
If i change the email on my profile and set the checkmark to keep my profile image and press submit, then the profile image is gone and Lansuite writes "Bitte fülle dieses Pflichtfeld aus." on the upload profile image button.

![image](https://github.com/lansuite/lansuite/assets/20731374/f6e3a300-8f67-4515-bf61-2ad43f9d2bd4)

M4LuZ:
What I've tried:
* Created new unprivileged user without profile picture - OK
* Uploaded profile picture - OK
* Changed Email address with existing profile picture - OK
* Removed Profile picture by deselecting checkbox - OK
* Changed Email address again without profile picture - OK
* Edited email of superadmin account without picture- OK
* Added profile picture to superadmin account - OK
* **Edited email of superadmin account with picture - ERROR**

Huh...

Pimmal:
Oh yeah i forgot i tried it on my admin profile.

PR #1107, opened 2024-04-14
Add /ext_inc/user_pics/ to ignored folders

M4LuZ:
### What is this PR doing?

adds `/ext_inc/user_pics/` to ignored folders.
I saw a uploaded picture pop up as modified while testing #1094
It's a flat directory structure, so no exclusion of subfolders needed. 

Fixes #

### Checklist

- [ ] ~`CHANGELOG.md` entry~
- [ ] ~Documentation update~


andygrunwald (approved):
There are still two files in this folder: https://github.com/lansuite/lansuite/tree/master/ext_inc/user_pics

I suggest
- to check if `[pic_.jpg](https://github.com/lansuite/lansuite/blob/master/ext_inc/user_pics/pic_.jpg)` is in use at all
- remove both if not in use
- add a `.gitkeep` file

Why removing those?
* We documented the folder structure here: https://lansuite.github.io/lansuite/docs/development/folder-structure/
* We (the system) should not ship code in `ext_inc` folder, because this folder should only contain user data (by removing those, we avoid instructions in the upgrade guide and make it a "single responsibility" folder)

What do you think @M4LuZ ?

andygrunwald:
@M4LuZ I added my comments into the code + added a part into the Upgrade guide.

PR #1114, opened 2024-04-29
Pw hashing rework

M4LuZ:
### What is this PR doing?

Recreation of PR #83 on current master with quite a good amount of further work done.
This is not yet complete, but close enough for a wider audience to have a look and provide feedback.

* Supports any algorithm from `pbkdf2`
* Sets `pbkdf2-sha512` as default
* Provides default config for all of them including fallback to a supported algorithm 
* Provides override of default config for hashing (if ever needed)
* Adds option to convert existing `md5` hashes to an intermediate custom format `md5-sha512`via update script with transparent changeover on login to `pbkdf2-sha512`, depends on a module framework rework or CLI updater to be fully usable
* Unit tests provided and passing (tests for method `hash`  and implementation code be done)
* Good amount of code simplifications (but some still to be done)
* Plus everything that the existing PR already did

refs #83 

### Checklist

- [ ] `CHANGELOG.md` entry
- [ ] Documentation update


andygrunwald:
@M4LuZ What is missing for this PR to move out of the draft status?

> This is not yet complete, but close enough for a wider audience to have a look and provide feedback.

M4LuZ:
Hi @andygrunwald ,

``modules/install/upgrade/2.0.php` is untested and has no means to be called currently.
I wanted to provide an easy mean to transfer existing md5 hashes to PBKDF2 as part of the upgrade (either by a proper module upgrade handling or a cli tool), which is a separate PR not yet done and sitting on my dev machine.

Also I was thinking about extending cases with `bcrypt` as that provides the similar properties as `argon2id`, but is part of the default PHP environment

But as for the original features and functionality this is feature-complete, tested and has unit test cases working

inc/Classes/Auth.php
```diff
@@ -332,7 +332,7 @@ public function login($email, $password, $show_confirmation = 1)
                 $func->log_event(t('Login fehlgeschlagen. Email (%1) nicht verifiziert', $user['email']), "2", "Authentifikation");
 
             // User login and wrong password?
-            } elseif ($user["user_login"] and $tmp_login_pass != $user["password"]) {
+            } elseif ($user["user_login"] and !PasswordHash::verify($password, $user["password"])) {
```
andygrunwald, line 335: Is the variable `$tmp_login_pass` still relevant?
In line 236, we still do md5 hashing, but continue to use the clear text password.

modules/clanmgr/Functions/CheckClanPW.php
```diff
@@ -8,8 +10,8 @@ function CheckClanPW($clanpw)
 {
     global $database;
 
-    $clan = $database->queryWithOnlyFirstRow("SELECT password FROM %prefix%clan WHERE clanid = ?", [$_GET['clanid']]);
-    if ($clan['password'] and $clan['password'] == md5($clanpw)) {
+    $clan = $db->qry_first("SELECT password FROM %prefix%clan WHERE clanid = %int%", $_GET['clanid']);
```
andygrunwald, line 13: Why is this query changed?
This can't work, as the global variable `$db` is not present.

My suggestion: Using the old database query code here
```suggestion
    $clan = $database->queryWithOnlyFirstRow("SELECT password FROM %prefix%clan WHERE clanid = ?", [$_GET['clanid']]);
```

modules/install/mod_settings/config.xml
```diff
@@ -174,6 +174,25 @@
...
     </entries>
   </typedefinition>
+  <typedefinition>
+    <head>
+      <name>passwordhashalgo</name>
+    </head>
+    <entries>
+      <entry>
+        <value>default</value>
+        <description>Standard</description>
```
andygrunwald, line 184: What is "Standard"?
Should we write "Standard (Algo)"?

modules/install/mod_settings/config.xml
```diff
@@ -310,6 +356,18 @@ eintagsmail.de
...
         <description>Email-Domain-Blacklist (Um Wegwerf-Mailadressen zu vermeiden)</description>
       </item>
+      <item>
+        <name>pwhash_algo</name>
+        <type>passwordhashalgo</type>
+        <default>default</default>
+        <description>Passwort Hashing Algorithmus</description>
+      </item>
+      <item>
+        <name>pwhash_algo_cfg</name>
```
andygrunwald, line 366: Same here, Maybe using `password_hashing_algorithm_config`

modules/usrmgr/add.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use LanSuite\PasswordHash;
```
andygrunwald, line 3: Why is this added, but not used?

modules/install/upgrade/2.0.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+//get users not having a hashed password
```
andygrunwald, line 3: How is this script called?
I (kind of) like the idea, but also think that people can just request a new password using the "password forget" function.

Primarily, if we don't have an update function that calls this script, this change would either add a new script that is unused or delay the PR.

inc/Classes/PasswordHash.php
```diff
@@ -0,0 +1,204 @@
+<?php
+
+namespace LanSuite;
+
+class PasswordHash
+{
+    //define default values to be used
```
andygrunwald, line 7:
```suggestion
```

The variable name says that this is the default configuration

inc/Classes/PasswordHash.php
```diff
@@ -0,0 +1,204 @@
...
+        }
+        //fallback to standard config if selection not supported on the system
+        return PasswordHash::PBKDF2_DEFAULT_CONFIG['default'];
+    }
+
+    private static function getAlgoCfg()
+    {
+        global $cfg;
+
+        $algo = self::getAlgo();
```
andygrunwald, line 184: Why is this function called, but the return value not used?

inc/Classes/PasswordHash.php
```diff
@@ -0,0 +1,204 @@
...
+     * Also returns a default config if selected algorithm has no config
+     * If selected algorithm is not supported on the system it is being overwritten with standard config
+     * @param string $algo The algorithm to receive the config for
+     * @return mixed[] Array of configuration options, currently only algorithm and iterations
+     */
+    public static function getDefaultAlgorithmConfig(string $algo = 'default') : array
+    {
+        if (PasswordHash::isAlgorithmSupported($algo)) {
+            // only consider Algorithm further if supported on the system
+            if (array_key_exists($algo, PasswordHash::PBKDF2_DEFAULT_CONFIG)){
```
andygrunwald, line 166: `$algo` is type string and we try to check if an array key exists. This can't work or am I missing something?

inc/Classes/PasswordHash.php
```diff
@@ -0,0 +1,204 @@
...
+     */
+    public static function hash(string $password, array $algo_cfg = null) :string
+    {
+        //use provided config or default
+        if ($algo_cfg == null) {
+            $algo_cfg = self::getAlgoCfg();
+        }
+        switch ($algo_cfg['algorithm']) {
+        case 'md5':
+            return md5($password);
```
andygrunwald, line 33: What is the rational still supporting md5?

tests/HashTest.php
```diff
@@ -0,0 +1,154 @@
...
+
+use function PHPUnit\Framework\assertEquals;
+
+class HashTest extends TestCase
+{
+
+    protected function setUp() : void
+    {
+        $GLOBALS['cfg'] = [];
+          error_reporting(E_ALL);
```
andygrunwald, line 17:
```suggestion
```
PHPUnit has its own error handling and we should use this: https://docs.phpunit.de/en/11.2/error-handling.html

tests/HashTest.php
```diff
@@ -0,0 +1,154 @@
...
+use ReflectionClass;
+
+use function PHPUnit\Framework\assertEquals;
+
+class HashTest extends TestCase
+{
+
+    protected function setUp() : void
+    {
+        $GLOBALS['cfg'] = [];
```
andygrunwald, line 16: Lets try to avoid globals in unit tests at all.
If this is needed, we should work on a bit of a refactoring.

tests/HashTest.php
```diff
@@ -0,0 +1,154 @@
...
+          error_reporting(E_ALL);
+
+        // the following test hashes will be used for all functions and tests, thus defined once and shared globally
+        $GLOBALS['hashTestArray'] =
```
andygrunwald, line 20: Lets use data providers for this: https://phpunit.de/manual/3.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers

tests/HashTest.php
```diff
@@ -0,0 +1,154 @@
...
+     */
+    public function testGetDefaultAlgorithmConfig()
+    {
+        $this->assertArrayHasKey('iterations', PasswordHash::GetDefaultAlgorithmConfig());
+    }
+
+    /*
+     * @covers PasswordHash::hash
+
+    public function testHash()
```
andygrunwald, line 57: Is this dead code and can be removed?

andygrunwald (changes requested):
I had a quick review and it seems we need to put a bit more work into this. @M4LuZ Let me know if you have time to do it or if you require some help.

Overall: I don't like the "global static" aspect of the class. This makes it pretty hard to re-use. I would rather love to see an object oriented approach with inheritance where each algorithm is its own sub class.

PR #1127, opened 2024-05-04
Import changes from old fork

M4LuZ:
### What is this PR doing?

There are multiple smaller fixes and improvements (and also some larger ones) that I've developed in the past in a fork.
As these are quite outdated I'll fix, refactor, improve as needed.
This is the first batch as foundation to then implement some larger and more interesting ones (as support for GiroCode QR-Code creation)

Fixes #none

### Checklist

- [x] `CHANGELOG.md` entry
- [x] Documentation update

### DIFF
--- a/inc/Classes/Display.php
+++ b/inc/Classes/Display.php
@@ -1278,9 +1278,11 @@ class Display
     }
 
     /**
-     * @param int $userid
-     * @param string $username
-     * @return string
+     * Adds user icon and username based on template ls_usericon.htm to output
+     *
+     * @param int $userid The numeric user id to fetch the icon for
+     * @param string $username The username to display, empty if not given
+     * @return string raw html for output
      * @throws \Exception
      * @throws \SmartyException
      */
@@ -1292,8 +1294,8 @@ class Display
             $username = '<i>System</i>';
         }
 
-        $smarty->assign('userid', htmlspecialchars($userid));
-        $smarty->assign('username', $username);
+        $smarty->assign('userid', $userid);
+        $smarty->assign('username', htmlspecialchars($username)); // username may contain special characters
         $smarty->assign('hint', t('Benutzerdetails aufrufen'));
 
         if (in_array($userid, $authentication->online_users)) {
--- a/inc/Classes/Display.php
+++ b/inc/Classes/Display.php
@@ -1292,7 +1292,7 @@ class Display
             $username = '<i>System</i>';
         }
 
-        $smarty->assign('userid', $userid);
+        $smarty->assign('userid', htmlspecialchars($userid));
         $smarty->assign('username', $username);
         $smarty->assign('hint', t('Benutzerdetails aufrufen'));
 
--- a/inc/Classes/Func.php
+++ b/inc/Classes/Func.php
@@ -1145,7 +1145,12 @@ class Func
             $replacementValues []= $auth['userid'];
         }
 
-        if (array_key_exists('userid', $auth)) {
+        if (array_key_exists('username', $auth)) {
+            $placeholderNames []= '%USERNAME%';
+            $replacementValues []= $auth['username'];
+        }
+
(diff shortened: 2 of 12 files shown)


inc/Classes/Display.php
```diff
@@ -1292,7 +1292,7 @@ public function FetchUserIcon($userid, $username = '')
             $username = '<i>System</i>';
         }
 
-        $smarty->assign('userid', $userid);
+        $smarty->assign('userid', htmlspecialchars($userid));
```
andygrunwald, line 1295: Isn't this always an integer?

inc/Classes/Func.php
```diff
@@ -1118,4 +1118,56 @@ public function isModActive($mod, &$caption = '')
...
+     */
+    public function replaceVariables($text) {
+        global $auth;
+
+        //initialize replacement array
+        $placeholderNames = [];
+        $replacementValues = [];
+
+            $placeholderNames []= '%TEST%';
+            $replacementValues []= 'Test OK!';
```
andygrunwald, line 1141: Is %TEST% really needed?

modules/party/Classes/Party.php
```diff
@@ -390,37 +399,51 @@ public function delete_usergroups($del_group, $set_group)
...
      * Returns the amount of users registered for a party.
-     * 
-     * @param int $party_id The ID of the party to calculate this for
+     *
+     * @param int $partyId The ID of the party to calculate this for (uses object value otherwise)
+     * @param
      * @return array Result array with elements "qty" and "paid"
     */
-    public function getGuestQty($party_id = NULL)
+    public function getGuestQty($partyId = null, $showOrga = null)
```
andygrunwald, line 410: Did we ever use the second parameter `$showOrga`?

website/docs/modules/Info2.md
```diff
@@ -0,0 +1,32 @@
...
+sidebar_position: 1
+---
+
+## Module Description
+
+This module provides the abilitity to create and manage multiple rich-text pages for user information pages
+
+## Configuration options
+
+TBD
```
andygrunwald, line 13: Is this still work in progress?

andygrunwald (changes requested):
I left a few comments

inc/Classes/Display.php
M4LuZ: This change was copied over by hand and I added it into the wrong line.
`$userid` is an integer, `$username` is (where it belongs) is a string with potentially html characters.
Will be changed.

inc/Classes/Func.php
M4LuZ: Nope. But it also does not hurt ;)
Fine to remove

website/docs/modules/Info2.md
M4LuZ: No I considered it really to be done in the future when also the other issues of the Info2 module are to be touched or we tackle generic module documentation.

I'm fine to remove the block if that is better?

andygrunwald: Yeah, lets remove it.

M4LuZ: I've decided to spend the few minutes to document it instead and noticed that option "extended toolbar" is not being used anywhere in the code, thus removed the option and documented the rest

modules/party/Classes/Party.php
M4LuZ: This _will_ be used, but is not at this point in time.
There is an config option to include registrations from admins/superadmins in the overview and I've added this to ensure that the function covers both cases. 
This is inteded to replace multiple calculations of similar type (if I remember correctly). 
If it turns out after that that it is not needed, then that's surely fine to remove

M4LuZ:
Should be fine to give it another look

andygrunwald (approved):
Looking good. Thank you!

PR #1141, opened 2024-05-29
Fix config caching, implement UID

### LINKED ISSUE #1139: Live und DEV system
I cloned lansuite into a dev subfolder, copied the DB and changed the URL.

Now I have the issue when I change user permissions in the DEV system, the user also gets changed in the live system. But both systems using different DB´s. Why?

Dev system
image

Live system
image

M4LuZ:
### What is this PR doing?

* Removes caching of global array `$config`
* Adds functionality to generate and store an uid to ensure uniqueness of cache pool per installation (new key in `config.cfg`)
* moves dependency on `Import`-Object in `Install` from class constructor to single function where it is used, as this causes issues when `$db` is not initialized due to another implicit dependency there

Fixes #1139

### Checklist

- [x] `CHANGELOG.md` entry
- [x] Documentation update


andygrunwald (approved):
@M4LuZ Good one! Thanks.

I extended this PR a bit, see https://github.com/lansuite/lansuite/pull/1141/commits/4eca718b149f24ad1c1a1cae37ee9c0afa6a273a

You removed the constructor parameter of the Install() class, but did not remove the usage of this in other modules.

I approve this PR and think it is good to merge.
Please do a quick check on https://github.com/lansuite/lansuite/pull/1141/commits/4eca718b149f24ad1c1a1cae37ee9c0afa6a273a - If you are fine with it, feel free to merge this PR.

M4LuZ:
Hi @andygrunwald ,

thanks for catching that. 
Did a bit of searching and testing, and apparently missed implicit checks for the constructor :/
Sure I'm fine with the modifications

PR #1154, opened 2024-05-30
Upgrade PHP minimum version from v8.1 to v8.2

andygrunwald:
### What is this PR doing?

This Pull Request is raising the minimum version of PHP to v8.2.
PHP v8.1 only receives security fixes.

Screenshot 2024-05-30 at 11 53 25

The idea is to be a little bit ahead and not being under pressure when we hit end of life for PHP v8.1 already.

PHP v8.2 was released 1 year, 5 months ago. Nearly every hoster has the option to run PHP 8.2 already.

None

### Checklist

- [X] `CHANGELOG.md` entry
- [X] Documentation update


andygrunwald:
Even the biggest frameworks in the PHP area like

* Symfony: https://github.com/symfony/http-foundation/blob/7.1/composer.json#L19
* PHPUnit: https://github.com/sebastianbergmann/phpunit/blob/main/composer.json#L25

only support active developed PHP versions.
What is that special on LanSuite on not doing so?
The challenge for us, as a _very_ small maintainer team is to support multiple versions. This makes it more complex, more error prone for testing (as our test suite is not good) and slows down development overall.

I think we can demand an active supported PHP version for people who are using such a system. I do believe the downsides of this are bigger as the upsides.

M4LuZ (approved):
I do see 8,2 being fairly well supported and haven't noticed any issues in that regard.
So I'm not blocking this, as you feel apparently a strong need to move forward.
But as said I find this too early to discontinue support without any need to do so. 

In regards to your arguments:
> Even the biggest frameworks in the PHP area like [...] only support active developed PHP versions.

That is not correct in three ways:
* Not _even_ these support, but _especially_ these big frameworks move forward early, as they have the need (and the resources) to adapt new versions quickly. Smaller projects move forward mostly when needed unless they have a very active / _enthusiastic_ development community.
* This is not a equal comparison as these ship releases, which we currently do not. 
Moving a dedicated development branch forward is also fine, but we do not have that.
* Symphony has active support for PHP 8.1 with 6.4 LTS up to 20208 apparently. 
They even still support with 5.4 LTS  a version that only requires PHP 7.2 till 2026. (See https://symfony.com/releases)
* PHPunit also has active support for a release that only requires 8.1  (See https://phpunit.de/supported-versions.html)

> The challenge for us, as a very small maintainer team is to support multiple versions

The issues we faced so far is not that we hit version specific bugs - which would definitely increase maintenance effort.
But instead we faced incompatibilities introduced after version increases in multiple cases.

Bumping the minimum version up does noting to improve that.

M4LuZ:
Merging this now as we had at least two successful parties with PHP 8.2 and dependencies require this now

PR #1172, opened 2024-06-10
Fix #1131: Restrict display of PHP warnings to (Super-)Admins

### LINKED ISSUE #1131: Restrict display of PHP warnings to (Super-)Admins
## Expected Behavior

End users should be informed if something did not work as expected, as they need to be aware that an operation did not finish successfully as they expected. 
Otherwise information not important or relevant for them should not be shown.
Administrators are users which should have an interest in warnings and further informations, so they should receive the related information

## Current Behavior

Warnings are shown to all users that both expose internal information and potentially confuse them, as there is a red message with some PHP warning.

## Possible Solution

* Set error_level to E_WARN early in the code.
* Once ` $cfg` and `$auth` are initialized switch error level depending on user level and debug setup
* If the user is an admin OR debugging is on then leave it on E_WARN
* Otherwise switch to E_ERROR

Find a place where a variable is not initialized, enjoy red warnings

## Context
This confuses users and leads to reports of _errors_ such as #1128, #1129, #1130

## Your Environment

* Version used:
* Operating System and version:
* Enabled features:

andygrunwald:
### What is this PR doing?

Surpresses PHP-Warnings if the user auth level is below Admin.

Fixes #1131

### Checklist

- [X] `CHANGELOG.md` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ We could not reconstruct _all_ changes, but we tried our best to make the most o
 - [Community] Created a discord community
 - [PDF] In-Module Documentation
 - [PDF] New options "align" and "border" to configure for a single PDF document
+- [System] Restrict display of PHP warnings to (Super-)Admins (#1131)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


M4LuZ (approved):
Thanks for that!

Issue #1191, opened 2024-07-05
Cannot install Lansuite


lukrei:
I cannot install Lansuite on debian 12 and MariaDB Version 10.11.6 and php 8.1.29 fpm.
During the installation the web installer asks me to create an admin account but admin account cannot be created.
During Step 6 setup wants me to create a admin user however this is not possible:
index.php?mod=install&action=wizard&step=6
In the apache error.log i get this message after i clicked on the Add button in step 6:
[proxy_fcgi:error] [pid 13776] [client 192.0.2.254:44170] AH01071: Got error 'PHP message: PHP Warning:  Undefined array key "design" in /var/www/lansuite/index.php on line 437', referer:

I also tried to change theme to OSx or something else but no difference.

Can someone please help me to install this version of lansuite?

M4LuZ:
Hi @lukrei ,

this is a known regression that was already raised as #1136 
Both that and the warning will be fixed with merge of PR #1193 
Our focus is currently on refactoring the codebase and keeping current installations running, but happy to get yet another user.

But: 
I would suggest that you give your other ticket #1192 an overhaul if you want any feedback or help on that

PR #1193, opened 2024-07-07
Fixed issue that no admin user is created in wizzard

### LINKED ISSUE #1136: Wizard - Creation of Admin user not possible
In the wizard on step 7 , I would like to create an admin user  but this fails. I cannot continue with the installation.

## Expected Behavior

The admin should be created based on the input and the process continues.

## Current Behavior
Bildschirmfoto 2024-05-21 um 15 57 58

The admin is not created and error messages appear

## Possible Solution

Implement method to create the user during the wizard or provide a workaround

```
    // Create Adminaccount
    case 7:
        //@TODO: Add the functionality here or remove the code
        break;
```

1. Fresh installation of landsite
2. Go through the wizard until admin creation (Step 7)
3. Try to create a new admin

## Context

Cannot install the application

## Your Environment

* Version used: master
* Operating System and version: Debian 12
* Enabled features: N/A

M4LuZ:
### What is this PR doing?

fixes issue caused by too accurate comparison in `wizzard.php` after implementation of `$database`.

Installer checked if return value from `$database->queryWithOnlyFirstRow` was not type equal to `false`.
But the new class returns an empty array when no entry is found, which does not compare to boolean false.
Thus the installer ran the update case, which did nothing as no entry existed.

This is fixed by just comparing against `$row`, as that covers everything that is not an actual resultset

While I was at it also the related queries have been transferred to use `$database` and also some warning from index.php that popped up as warning.
Also transparent transfer after user creation to next step was implemented.

Fixes #1136, #1191

### Checklist

- [x] `CHANGELOG.md` entry
- [ ] ~Documentation update~

### DIFF
--- a/modules/install/wizard.php
+++ b/modules/install/wizard.php
@@ -27,25 +27,24 @@ switch ($step) {
             $row = $database->queryWithOnlyFirstRow("SELECT email FROM %prefix%user WHERE email = ?", [$_POST["email"]]);
 
             // If found, update password
-            if ($row !== false) {
-                $db->qry(
-                    "UPDATE %prefix%user SET password = %string%, type = '3' WHERE email=%string%",
-                    md5($_POST["password"]),
-                    $_POST["email"]
-                );
-            // If not found, insert
-            } else {
-                $db->qry(
-                    "INSERT INTO %prefix%user SET username = 'ADMIN', firstname = 'ADMIN', name = 'ADMIN', email=%string%, password = %string%, type = '3', lastlogin = NOW(), comment = '', birthday = NULL, signature = ''",
-                    $_POST["email"],
-                    md5($_POST["password"])
-                );
-                $userid = $db->insert_id();
-            }
-
-            $authentication = new \LanSuite\Auth($database, $request);
-            $authentication->login($_POST["email"], $_POST["password"]);
+        if ($row) {
+            $database->query(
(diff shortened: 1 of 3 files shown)


modules/install/wizard.php
```diff
@@ -27,25 +27,24 @@
...
-                    $_POST["email"]
-                );
-            // If not found, insert
-            } else {
-                $db->qry(
-                    "INSERT INTO %prefix%user SET username = 'ADMIN', firstname = 'ADMIN', name = 'ADMIN', email=%string%, password = %string%, type = '3', lastlogin = NOW(), comment = '', birthday = NULL, signature = ''",
-                    $_POST["email"],
-                    md5($_POST["password"])
-                );
-                $userid = $db->insert_id();
```
andygrunwald, line 43: Are we 100% sure that `$userid` is not used in an include context?
Because this is not retrieved in the new code.

andygrunwald (approved):
Looks good (only one small comment, but I did a quick check and was not able to find it)

modules/install/wizard.php
M4LuZ: Hi Andy,

I had the same thought, but there was no other occurrence in the file and multiple end-to-end runs of the installer showed no warnings or issues.

My guess is that this was an leftover from when login functionality was transferred to `Auth->login`

PR #1314, opened 2025-03-18
Pw hash include argon2

hnxfirefly:

### DIFF
--- a/inc/Classes/Auth.php
+++ b/inc/Classes/Auth.php
@@ -332,7 +332,7 @@ class Auth
                 $func->log_event(t('Login fehlgeschlagen. Email (%1) nicht verifiziert', $user['email']), "2", "Authentifikation");
 
             // User login and wrong password?
-            } elseif ($user["user_login"] and $tmp_login_pass != $user["password"]) {
+            } elseif ($user["user_login"] and !PasswordHash::verify($password, $user["password"])) {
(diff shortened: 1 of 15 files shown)


M4LuZ:
While I am very happy to see a new PR I do not see the point of this.
There is a pending PR for a multi-hash implementation that should move forward shortly once dependencies  #870 & #1178 are resolved (yes, I know how that sounds, but really!) 
It is trivial to add argon2id there if you really want.
That would be better than to do the same again without taking over the improvements done (multi-hash support, documentation, unit tests)

But the main point why Argon2id was not implemented there is that it is still not part of default php installations - as stated by the other PR. Not sure where you get a different impression from?

See  [related RFC](https://wiki.php.net/rfc/argon2_password_hash) 

> "Per the discussion on https://github.com/php/php-src/pull/1997, this feature will be optionally available via the --with-argon2 configure flag." 

Or the PHP documentation itself: 
> "[PASSWORD_ARGON2ID](https://www.php.net/manual/en/password.constants.php#constant.password-argon2id) - Use the Argon2id hashing algorithm to create the hash. This algorithm is only available if PHP has been compiled with Argon2 support."

In PHP 8.4, the cost parameter of the password_default/password_bcrypt algorithm has changed from 10 to 12. But it is still bcrypt by default.

